PR #1, opened 2018-05-28
add "-DWITH_ASAN=ON -DWITH_UBSAN=ON" option

delgod:


laurynas-biveinis (changes requested):
Could you please add -DWITH_ASAN=ON -DWITH_ASAN_SCOPE=ON -DWITH_UBSAN=ON too?

delgod:
@laurynas-biveinis, I have added one more option

laurynas-biveinis (approved):

PR #5, opened 2018-06-12
add default value for GIT_REPO variable

delgod:


laurynas-biveinis (approved):
LGTM. What about 5.5 and 5.6?

delgod:
@laurynas-biveinis I am going to rebase 5.5 and 5.6 branches on this commit, so changes will appear in all branches.
is it ok?

laurynas-biveinis:
Sure, I don't see any problems with that

PR #7, opened 2018-06-13
use release builds compilers on CentOS 6 32bit/64bit

delgod:


laurynas-biveinis (commented):
How does this compare to
- old Jenkins
- how release packages are built?

delgod:
@laurynas-biveinis, right now - these jobs don't have relation to release jobs.
release jobs use build scripts which are located in percona-server repo and I am ok to don't modify build scripts inside percona-server repo, so no relations right now.

anyway, it is very important for us to get migration done without adding any instability for old working jobs.
we don't have plans to migrate release jobs to AWS build server.

laurynas-biveinis:
I am not asking to merge the scripts right now - I am asking about the net effect, what versions of GCC do the current release scripts and old Jenkins use?

delgod:
32bit CentOS 6 - 4.4.7 (OS default version)
64bit CentOS 6 - 4.8.2 (devtoolset-2 version)

devtoolset-4 has 5.3.1 version
devtoolset-6 has 6.3.1 version

laurynas-biveinis:
@delgod, is that old Jenkins or release scripts?

delgod:
we have release jobs only on old Jenkins.

param/trunk and release jobs on old Jenkins have the following versions:
32bit CentOS 6 - 4.4.7 (OS default version)
64bit CentOS 6 - 4.8.2 (devtoolset-2 version)

laurynas-biveinis:
@delgod, I believe we should match then, by changing either Jenkins testing compiler version, release package build compiler version, or both. We have seen compiler-version-dependent behavior in the past and we have to test what we ship.

delgod:
@laurynas-biveinis, please review again.
now PS build server has the same versions of compilers in tests and in release builds.

laurynas-biveinis (approved):

Issue #8, opened 2018-06-15
MTR tests should run with a non-root user


laurynas-biveinis:
Right now, e.g. https://ps.cd.percona.com/job/percona-server-5.7-pipeline/1078/consoleText:
`[ERROR] Fatal error: Please read "Security" section of the manual to find out how to run mysqld as root!`

Please also add a whoami call for diagnostics

delgod:
please review https://github.com/Percona-Lab/ps-build/pull/15

delgod:
fixed in fact
```
main.drop-no_root                        w2 [ pass ]      4
```
MTR log - https://s3.us-east-2.amazonaws.com/ps-build-cache/jenkins-percona-server-5.6-pipeline-648/mtr.output.gz

PR #15, opened 2018-06-20
#8 run build/test as non-root user

delgod:


docker/run-build
```diff
@@ -35,11 +35,11 @@ docker run --rm \
 
     mkdir  /tmp/results
     cp -r /tmp/source_downloads /tmp/results/source_downloads
+    sudo chown mysql:mysql /tmp/ps/mysql-test/collections /tmp/ps/storage/rocksdb/rocksdb/util
```
delgod, line 38: this line needed because we modify source tree during cmake/make:
- https://jira.percona.com/browse/PS-4536
- https://jira.percona.com/browse/PS-4537

laurynas-biveinis (approved):

docker/run-build
laurynas-biveinis: This line is going to stay even if the above are fixed due to need to build upstream server and old versions

PR #42, opened 2018-08-30
JEN-894 centos 6 add asan

delgod:


laurynas-biveinis (approved):

docker/install-deps
```diff
@@ -43,8 +43,9 @@ if [ -f /usr/bin/yum ]; then
 		gpgkey=http://linuxsoft.cern.ch/cern/slc6X/\$basearch/RPM-GPG-KEY-cern
 		gpgcheck=1
 	EOF
-        PKGLIST=${PKGLIST/ gcc-c++/ devtoolset-2-gcc-c++ devtoolset-2-binutils devtoolset-6-gcc-c++ devtoolset-6-binutils}
+        PKGLIST=${PKGLIST/ gcc-c++/ devtoolset-2-gcc-c++ devtoolset-6-gcc-c++}
```
abychko, line 46: devtoolset-2-gcc-c++ and devtoolset-6-gcc-c++ in the same time?

delgod: yes, we use the same docker images for all branches (which are prepared by `prepare-docker` command).
so it means that image should contain requirements for all PS versions.
So devtoolset-2 is needed for PS 5.7, and devtoolset-6 for PS 8.0

laurynas-biveinis:
Still not working: https://ps.cd.percona.com/job/percona-server-8.0-pipeline/597/consoleText

PR #49, opened 2018-09-13
JEN-910 Changes for 8.0 Valgrind job

abychko:
[*] changed MTR --parallel to cores/2 for valgrind builds
[*] increased build timeout to 15h
[+] added new platforms to the matrix

### DIFF
--- a/jenkins/pipeline.groovy
+++ b/jenkins/pipeline.groovy
@@ -1,3 +1,12 @@
+pipeline_timeout = 10
+
+if (
+    (params.ANALYZER_OPTS.contains('-DWITH_VALGRIND=ON'))   ||
+    (params.ANALYZER_OPTS.contains('-DWITH_ASAN=ON'))       ||
+    (params.ANALYZER_OPTS.contains('-DWITH_UBSAN=ON'))
+    )
+    {  pipeline_timeout = 15 }
+
 pipeline {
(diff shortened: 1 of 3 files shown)


delgod:
why we need two times less threads in all params jobs?

abychko:
> why we need two times less threads in all params jobs?

only if [[ "${ANALYZER_OPTS}" == *WITH_VALGRIND=ON* ]]

jenkins/pipeline.groovy
```diff
@@ -105,7 +105,7 @@ pipeline {
         compressBuildLog()
         skipDefaultCheckout()
         skipStagesAfterUnstable()
-        timeout(time: 10, unit: 'HOURS')
+        timeout(time: 15, unit: 'HOURS')
```
laurynas-biveinis: This is global right? Possible to limit to Valgrind jobs only?

abychko: yes, should be possible to use variable with default value

laurynas-biveinis: On the other hand, looking at sanitizer run results, please set 15 h there too

laurynas-biveinis (approved):
Looks good, let's take a look at a Valgrind test job: start and abort it when it gets to MTR

PR #51, opened 2018-09-14
JEN-911 Misleading typo in parameter description: "--only-big-tests"

abychko:
[*] fixed typo


laurynas-biveinis (changes requested):
This option does not exist in 5.6 and its mentions should be removed instead

abychko:
deleted, please review

laurynas-biveinis (approved):

PR #54, opened 2018-09-18
JEN-913 prevent wrong version run

abychko:


jenkins/pipeline.groovy
```diff
@@ -85,6 +85,21 @@ pipeline {
                 }
 
                 sh 'echo Prepare: \$(date -u "+%s")'
+                echo 'Checking Percona Server branch version, JEN-913 prevent wrong version run'
+                sh '''
+                    MY_BRANCH_BASE_MAJOR=5
+                    MY_BRANCH_BASE_MINOR=5
+                    RAW_VERSION_LINK=$(echo ${GIT_REPO%.git} | sed -e "s:github.com:raw.githubusercontent.com:g")
```
delgod, line 92: why do we need to fetch it?
I think we can check version file content after `./local/checkout` command

abychko: the idea is to terminate the job immediately and don't do checkout in case of version mismatch

jenkins/pipeline.groovy
```diff
@@ -85,6 +85,21 @@ pipeline {
...
+                    wget ${RAW_VERSION_LINK} -O ${WORKSPACE}/VERSION-${BUILD_NUMBER}
```
laurynas-biveinis: What does RAW_VERSION_LINK resolve to on a sample run?

abychko: https://ps.cd.percona.com/job/test-percona-server-5.7-pipeline/8/console

laurynas-biveinis (approved):
LGTM if looks good to @delgod

abychko:
wrong version https://ps.cd.percona.com/view/5.5/job/test-percona-server-5.5-pipeline/6/
right version https://ps.cd.percona.com/view/5.5/job/test-percona-server-5.5-pipeline/5/

laurynas-biveinis (approved):

PR #57, opened 2018-09-18
JEN-913 8.0 prevent wrong version run

abychko:


laurynas-biveinis (changes requested):
For the highest version, which is 8.0 currently, only forbid building lower versions. If 8.1 or 9.0 comes along, we'd try building it with 8.0 jobs at first

abychko:
higher versions are allowed

abychko:
right version https://ps.cd.percona.com/view/5.5/job/test-percona-server-8.0-pipeline/11/
wrong version https://ps.cd.percona.com/view/5.5/job/test-percona-server-8.0-pipeline/12/

higher version is not tested because we have no 8+ branches in our repo

laurynas-biveinis (approved):

laurynas-biveinis (approved):

PR #63, opened 2018-10-05
Jen 926 8.0

abychko:


abychko:
https://ps.cd.percona.com/view/8.0/job/test-percona-server-8.0-pipeline/18/

laurynas-biveinis (changes requested):
The awk invocation is repeated three times and is not trivial. Can it be extracted to shell something in order to avoid two copy-pastes?

abychko:
yes, it can be extracted, but please make it separate task. need to fight variables dereferension in single/double quotes

laurynas-biveinis:
OK, can you please file a JEN issue yourself?

laurynas-biveinis (approved):

PR #64, opened 2018-10-08
JEN-927 8.0

abychko:
merge 3 awk calls in local/test-binary
[+] added separate for loop to process .output files


abychko:
test is running: https://ps.cd.percona.com/view/8.0/job/test-percona-server-8.0-pipeline/19/

PR #68, opened 2018-10-17
JEN-936

abychko:
Percona server build failures at git clean
[+] added chown on ps source dir

### DIFF
--- a/local/checkout
+++ b/local/checkout
@@ -18,6 +18,7 @@ if [ ! -d "${ROOT_DIR}" ]; then
     git clone "${GIT_REPO:-https://github.com/percona/percona-server}" "${ROOT_DIR}"
 fi
 
+sudo chown -R $USER $ROOT_DIR
 pushd $ROOT_DIR
     git reset --hard
     git clean -xdf


laurynas-biveinis (commented):
Please show testing evidence for @dutow
Why git reset and git clean changed places?

abychko:
re: testing evidence. I don't know how to reproduce build on reused VM or docker image to have broken permissions in the source tree. I can ran builds as usual, but they can't give us any proof.

re: git. it's better to clean before reset because some files may exist or not exist in different branches or revisions.

laurynas-biveinis:
> re: testing evidence. I don't know how to reproduce build on reused VM or docker image to have broken permissions in the source tree. I can ran builds as usual, but they can't give us any proof.

ack, please discuss with @inikep on Slack/JIRA

> re: git. it's better to clean before reset because some files may exist or not exist in different branches or revisions.

This is exactly why I think reset first is better - in the case of branch change it will determine which files should be removed by a later clean. In any case, is it directly related to permission fix? If not, please PR separately

dutow:
@abychko Looking at the build history for this morning, first I started a build, then aborted it as I forgot to push an update - maybe that could be the cause?
After that, I tried starting a new build several times, but it always failed within minutes. Now that a few hours passed, it works again.

abychko:
Yes, the reason is reusing of dirty image. I believe we should fix it anyway

abychko:
changed back order of git clean & reset

abychko:
> > re: testing evidence. I don't know how to reproduce build on reused VM or docker image to have broken permissions in the source tree. I can ran builds as usual, but they can't give us any proof.
> 
> ack, please discuss with @inikep on Slack/JIRA

sorry, discuss what? build on dirty images?

laurynas-biveinis:
> sorry, discuss what? build on dirty images?

It seems you already have discussed above

laurynas-biveinis:
ad0edf1 does not do what it says it should. And since this is a non-trunk branch, a better practice is to remove commit and push with --force-with-lease rather than add reverting commit on the top

abychko:
I can squash and merge commits as one commit

abychko:
done according to your comments

delgod:
btw I think it is better to run `sudo chown` before `local/checkout` script in pipeline https://github.com/Percona-Lab/ps-build/blob/5.7/jenkins/pipeline.groovy#L157
not inside `local/checkout` script which can run without root on developer machine.

abychko:
ok, makes sense

abychko:
issue is reproducible locally, without Jenkins. so, chown call should be in script, not in pipeline

laurynas-biveinis (approved):
OK if @delgod agrees with the general approach and if testing evidence is provided

laurynas-biveinis:
Should this be closed?

abychko:
closed or merged & closed?

laurynas-biveinis:
I don't know. It's opened for six weeks now

PR #83, opened 2018-12-13
JEN-984

abychko:
Remove worker number from the XML testcase name


abychko:
started https://ps2.cd.percona.com/view/5.6/job/test-percona-server-5.6-pipeline/2/

laurynas-biveinis (approved):
OK if testing OK

PR #87, opened 2018-12-18
8.0: switch to -DWITH_SYSTEM=libs

abychko:
[-] deleted every occurence of WITH_SSL from Jenkins GUI and pipeline
[-] replaced WITH_SSL with WITH_SYSTEM=libs in build scripts
[-] deleted WITH_ZLIB

### DIFF
--- a/local/build-binary
+++ b/local/build-binary
@@ -12,7 +12,6 @@ COMPILER=${COMPILER:-default}
 CMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE:-RelWithDebInfo}
 WITH_TOKUDB=${WITH_TOKUDB:-ON}
 WITH_ROCKSDB=${WITH_ROCKSDB:-ON}
-WITH_SSL=${WITH_SSL:-system}
 ANALYZER_OPTS=${ANALYZER_OPTS:-}
 MAKE_OPTS=${MAKE_OPTS:--j$(grep -c ^processor /proc/cpuinfo)}
 CMAKE_OPTS=${CMAKE_OPTS:-}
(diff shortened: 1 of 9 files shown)


local/build-binary
```diff
@@ -151,8 +150,7 @@ pushd ${WORKDIR}
         -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} \
         -DWITH_TOKUDB=${WITH_TOKUDB} \
         -DWITH_ROCKSDB=${WITH_ROCKSDB} \
-        -DWITH_SSL=${WITH_SSL} \
-        -DWITH_MECAB=${WITH_MECAB} \
```
laurynas-biveinis: MECAB should stay, because system libs are documented as "This option serves as an “umbrella” option to set the system value of any of the following CMake options that are not set explicitly: WITH_CURL, WITH_EDITLINE, WITH_ICU, WITH_LIBEVENT, WITH_LZ4, WITH_LZMA, WITH_PROTOBUF, WITH_RE2, WITH_SSL, WITH_ZLIB." - mecab is not there

laurynas-biveinis (changes requested):

local/build-binary
abychko: reverted that line

laurynas-biveinis (approved):
OK if testing OK

abychko:
started https://ps2.cd.percona.com/view/8.0/job/test-percona-server-8.0-pipeline/18/console

laurynas-biveinis:
This needs full testing on all slaves to catch any missing libs there or the build will break - has this been done?

laurynas-biveinis:
Oh, it's not working anyway, please revert and re-PR: it's not -DWITH_SYSTEM=libs, it's -DWITH_SYSTEM_LIBS=ON

abychko:
changed PR to include -DWITH_SYSTEM_LIBS=ON

PR #95, opened 2018-12-27
JEN-999 8.0 Increase Jenkins Valgrind job timeout to 6 days

abychko:
increased timeout for valgrind and asan builds


abychko:
also slightly increased ASAN, we have it timed out too

PR #98, opened 2019-01-15
JEN-936 5.6

abychko:
[*] run 'git clean' under sudo because cmake creates some files under
root account (default in docker)

### DIFF
--- a/local/checkout
+++ b/local/checkout
@@ -26,7 +26,7 @@ pushd $ROOT_DIR
     fi
 
     git reset --hard
-    git clean -xdf
+    sudo git clean -xdf
 
     if [ -n "${BRANCH}" ]; then
         git checkout "${BRANCH}"


abychko:
started https://ps2.cd.percona.com/view/5.6/job/test-percona-server-5.6-pipeline/4/console

abychko:
git clean is OK under sudo on the same VM https://ps2.cd.percona.com/view/5.6/job/test-percona-server-5.6-pipeline/5/console

dutow:
Any git command has a chance to run git gc (or in the future: possibly other commands), which modifies the metadata directory. We also can't guarantee that git won't add metadata modification changes to this command later. So while this change most of the time will fix the issue (removal of the temporary files from the directory), it also has the possibility of breaking things even more (adding root owned files to the .git directory).

Good fixes for this would be:

* to chown everything to the current user before cleaning stuff. But this also requires sudu, and the fact that our jenkins user has passwordless sudo right on the build machine is a *huge* security risk
* so the real solution would be modifying the docker script to use the correct uid and gid - there are several ways to do that

PR #100, opened 2019-01-15
JEN-936 8.0

abychko:
[*] run 'git clean' under sudo because cmake creates some files under
root account (default in docker)


delgod:
I believe it is better to run this command with sudo on Jenkins level

abychko:
looks good https://ps2.cd.percona.com/view/8.0/job/test-percona-server-8.0-pipeline/28/console

PR #102, opened 2019-01-15
Jen 936 5.6

abychko:
[*] run chown on source directory if it exist. done that at script level because original issue is from script


abychko:
https://ps2.cd.percona.com/view/5.6/job/test-percona-server-5.6-pipeline/6/console with chown

PR #111, opened 2019-01-22
JEN-976

abychko:
Remove 32-bit build support remnants from 8.0

note: build script for docker images must contain i386/centos/6 because
we use it for building images for all PS versions


abychko:
success https://ps2.cd.percona.com/view/8.0/job/test-percona-server-8.0-pipeline/31/console

PR #130, opened 2019-02-06
JEN-1025 install devtoolset-7 for centos6

hors:


abychko (commented):
seems we need to update i386/centos:6 and figure out why it's outdated. 64bit image built with the same code, so I don't think we need changes here

PR #141, opened 2019-02-18
JEN-1025 use 'devtoolset-2' for 'centos6-i386'

hors:


EvgeniyPatlan (approved):

local/build-binary
```diff
@@ -113,6 +113,10 @@ if [[ -f /opt/rh/devtoolset-7/enable ]]; then
     source /opt/rh/devtoolset-7/enable
 fi
 
+# Centos 6/i386
```
laurynas-biveinis, line 116: The above block enables devtoolset-7, better to rewrite it as 
if 386 && -f devtoolset-2
enable -2
elif -f devtoolset-7
enable -7
fi
, same as in https://github.com/Percona-Lab/ps-build/pull/145

PR #148, opened 2019-02-21
JEN-1022 add separate jobs for ps fb

hors:


EvgeniyPatlan (approved):
LGTM

jenkins/param-fb.yml
```diff
@@ -0,0 +1,121 @@
+- job:
+    name: percona-server-8.0-fb-param
```
inikep: This job has nothing in common with Percona Server. It should be `fb-mysql-server-8.0` or more generic `mysql-server-8.0` (for testing both MySQL and FB-MySQL).

hors: ok, I see your point. I have fixed it.

jenkins/pipeline-fb.groovy
```diff
@@ -0,0 +1,229 @@
...
+if (params.ANALYZER_OPTS.contains('-DWITH_VALGRIND=ON'))
+    { pipeline_timeout = 144 }
+
+if (
+    ((params.ANALYZER_OPTS.contains('-DWITH_ASAN=ON')) &&
+    (params.ANALYZER_OPTS.contains('-DWITH_ASAN_SCOPE=ON')) &&
+    (params.ANALYZER_OPTS.contains('-DWITH_UBSAN=ON'))) ||
+    ((params.MTR_ARGS.contains('--big-test')) || (params.MTR_ARGS.contains('--only-big-test')))
+    ) {
+        LABEL = 'docker-32gb'
```
percona-ysorokin: This `LABEL = 'docker-32gb'` is probably not needed here as we always use `docker-32gb`
```suggestion
```

hors: yes, you are right we can remove it

percona-ysorokin (approved):
LGTM

PR #167, opened 2019-03-25
JEN-1066 add ubuntu 'disco' to PS jenkinses

hors:


jenkins/pipeline.groovy
```diff
@@ -25,7 +25,7 @@ pipeline {
             description: 'Tag/Branch for Percona-TokuBackup repository',
             name: 'TOKUBACKUP_BRANCH')
         choice(
-            choices: 'ubuntu:artful\ncentos:6\ncentos:7\ni386/centos:6\nubuntu:trusty\nubuntu:xenial\nubuntu:bionic\nubuntu:cosmic\ndebian:jessie\ndebian:stretch',
+            choices: 'ubuntu:artful\ncentos:6\ncentos:7\ni386/centos:6\nubuntu:trusty\nubuntu:xenial\nubuntu:bionic\nubuntu:cosmic\nubuntu:disco\ndebian:jessie\ndebian:stretch',
```
EvgeniyPatlan: We should remove artuful as it is unsupported OS for now so no need to run tests on it

jenkins/trunk.yml
```diff
@@ -145,6 +145,7 @@
           - ubuntu:artful
```
EvgeniyPatlan: the same

EvgeniyPatlan (changes requested):

jenkins/pipeline.groovy
hors: done

jenkins/trunk.yml
hors: done

EvgeniyPatlan (approved):

PR #169, opened 2019-03-26
JEN-957 fix trunk-fips job

hors:
the same changes which were made under tasks JEN-1057  for 5.6 5.7


EvgeniyPatlan (approved):
lgtm if tests are ok

PR #170, opened 2019-03-28
JEN-1065 do not pass '-DWITH_KEYRING_VAULT=ON' for Centos6

hors:


hors:
https://ps3.cd.percona.com/view/8.0/job/percona-server-8.0-pipeline/500/#showFailuresLink

local/test-binary
```diff
@@ -19,6 +19,12 @@ rm -fr ${WORKDIR_ABS}/PS
 mkdir -p ${WORKDIR_ABS}/PS/sql
 tar -C ${WORKDIR_ABS}/PS --strip-components=1 -zxpf $(ls $WORKDIR_ABS/*.tar.gz | head -1)
 cd ${WORKDIR_ABS}/PS/mysql-test
+
+if [[ "$(lsb_release -d -s)" = *"CentOS release 6."* ]]; then
+    mkdir ./suite/keyring_vault
+    rsync -aHv /tmp/ps/plugin/keyring_vault/tests/mtr/ ./suite/keyring_vault/
```
laurynas-biveinis, line 25: Wait, so running the testsuite will still be broken outside of Jenkins in CentOS 6. The installation of keyring_vault tests must happen in the server packaging scripts, in the PS repo

laurynas-biveinis (changes requested):

local/test-binary
hors: @EvgeniyPatlan Please have a look.

hors: > Wait, so running the testsuite will still be broken outside of Jenkins in CentOS 6. The installation of keyring_vault tests must happen in the server packaging scripts, in the PS repo

I think we need to fix it in two places because we do not use script from PS repo for building PS inside  of Jenkins.

hors: @laurynas-biveinis The task was created: https://jira.percona.com/browse/ENG-125

laurynas-biveinis: If it's fixed in CMake so that "make install" always install keyring_vault tests, that should be the only place needing fixing. Please create a PS ticket for this (or move ENG-125)

hors: https://jira.percona.com/browse/PS-5537 done

PR #172, opened 2019-03-29
JEN-1069 create '/bin/python' symlink for RHEL8

hors:
https://ps3.cd.percona.com/view/5.7/job/percona-server-5.7-pipeline/1397/consoleFull - ok

### DIFF
--- a/docker/install-deps
+++ b/docker/install-deps
@@ -83,6 +83,12 @@ if [ -f /usr/bin/yum ]; then
         sleep 1
     done
 
+    if [ "x${RHVER}" = 'x8' ]; then
+        if [ -f '/bin/python2' -a ! -f '/bin/python' ]; then
+            ln -s -f /bin/python2 /bin/python
+        fi
+    fi
+
     # FIXME. WORKAROUND. it's needed for libeatmydata.
     # FIXME. it's already installed, but some files are missing for some reason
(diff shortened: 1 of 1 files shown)


laurynas-biveinis (changes requested):
Such fix means that our Python scripts are not runnable out of the box on RHEL8

hors:
> Such fix means that our Python scripts are not runnable out of the box on RHEL8

It is just a temporary fix for our tests till scrips are fixed. Using python2 scripts work fine. When I tried  to use python3 I had a lot of errors:
https://ps3.cd.percona.com/view/5.7/job/percona-server-5.7-pipeline/1422/consoleFull

According to the following information https://developers.redhat.com/blog/2018/11/14/python-in-rhel-8/ rhel8 will have python3 installed by default so it is more preferable to use python3 for ours scripts.

laurynas-biveinis:
Such temporary fixes have a tendency to stay permanent and we cannot say that our product supports RHEL8 anyway, while they are present, because nobody else can repeat the build/test without hacking their system first.

hors:
> Such temporary fixes have a tendency to stay permanent and we cannot say that our product supports RHEL8 anyway, while they are present, because nobody else can repeat the build/test without hacking their system first.

Ok, thank you for explanation. I am closing this review in such a case.

PR #173, opened 2019-03-29
JEN-1065 do not pass '-DWITH_KEYRING_VAULT=ON' for Centos6

hors:


delgod (approved):

laurynas-biveinis (commented):
This cannot be merged until keyring_vault test installation gets fixed, or CentOS 6 test runs will fail completely. Or do you have a test run showing otherwise?

hors:
> This cannot be merged until keyring_vault test installation gets fixed, or CentOS 6 test runs will fail completely. Or do you have a test run showing otherwise?

@laurynas-biveinis I totally agree with you and I mentioned the same in PS-5537.

PR #175, opened 2019-04-02
JEN-957 enable cron for trunk job

hors:


laurynas-biveinis:
What does this schedule say?

laurynas-biveinis (changes requested):
Friday's run is too late for its week. Let's move the schedule one day back (Sunday to Thursday 11PM), because then we have fresh results on Monday

hors:
> Friday's run is too late for its week. Let's move the schedule one day back (Sunday to Thursday 11PM), because then we have fresh results on Monday

ok, done

delgod (approved):

PR #176, opened 2019-04-02
JEN-1047 add timeout per stage and increase global timeout

hors:


jenkins/pipeline.groovy
```diff
@@ -115,122 +115,128 @@ pipeline {
...
-                retry(3) {
-                git branch: '8.0', url: 'https://github.com/Percona-Lab/ps-build'
-                    withCredentials([string(credentialsId: 'MTR_VAULT_TOKEN', variable: 'MTR_VAULT_TOKEN')]) {
-                        sh '''
-                            sudo git reset --hard
-                            sudo git clean -xdf
-                            rm -rf sources/results
-                            sudo git -C sources reset --hard || :
-                            sudo git -C sources clean -xdf   || :
+                timeout(time: pipeline_timeout, unit: 'HOURS')  {
```
delgod, line 209: with `-DWITH_VALGRIND` pipeline_timeout = 144 but it has no sense because the general timeout is 3 days

hors: thank you, fixed

delgod (approved):

PR #181, opened 2019-04-15
JEN-1078 decrease value for '--testcase-timeout' option

hors:


laurynas-biveinis (approved):
Please also change 5.6 and 5.7

hors:
> Please also change 5.6 and 5.7

Ok, I will

PR #186, opened 2019-04-23
JEN-1084 do not allow setting/changing parameters for trunk jobs

hors:


jenkins/trunk.yml
```diff
@@ -17,115 +18,10 @@
...
-        name: MTR_REPEAT
-        default: "1"
-        description: Run each test N number of times, --repeat=N
-    - choice:
-        name: LABEL
-        choices:
-        - docker
-        - docker-32gb
-        description: Run build on specified instance type
+        description: "Do not change the value of this patameter. This is a trunk job and it should not be run manually."
```
laurynas-biveinis:
```suggestion
        description: "Do not change the value of this parameter. This is a trunk job."
```

jenkins/trunk.yml
```diff
@@ -3,6 +3,7 @@
     project-type: matrix
     defaults: global
     description: |
+        Do not run this job manually!
```
laurynas-biveinis: It's OK to run trunk jobs manually, it's not OK to change params
```suggestion
```

laurynas-biveinis (changes requested):

jenkins/trunk.yml
hors: Thank you, fixed

jenkins/trunk.yml
hors: I have corrected it.

delgod (approved):

PR #190, opened 2019-04-24
JEN-1085 use docker-32gb worker for PS 5.6

hors:


delgod (approved):
just keep in mind - param build now 32% more expensive

laurynas-biveinis:
docker-32gb in 5.6 is only for --big-test (which trunk jobs have by default, param jobs don't but it's allowed to add it there). If we can parameterize on its presence, param build would become cheaper again

PR #193, opened 2019-04-25
JEN-1084 change the value for --testcase-timeout option

hors:
* in case of 'big-test' we need to use bigger value for
      '--testcase-timeout' option to avoid timeout issue in
      tests


hors:
@laurynas-biveinis Should I do the same for 5.7(because now we do not have timeout issue in 5.7)?

hors:
https://ps3.cd.percona.com/view/5.6/job/percona-server-5.6-pipeline/1259/
https://ps3.cd.percona.com/view/5.6/job/percona-server-5.6-pipeline/1257/

laurynas-biveinis (approved):

laurynas-biveinis:
5.7 looks OK, although hard to reconcile while Valgrind multiplier here is 15 and there is 2. We can tweak when needed

PR #194, opened 2019-04-25
JEN-1084 change the value for --testcase-timeout option

hors:
* in case of 'only-big-test' we need to use bigger value for
      '--testcase-timeout' option to avoid timeout issue in
      tests

### DIFF
--- a/local/test-binary
+++ b/local/test-binary
@@ -113,6 +113,11 @@ if [[ ! -e /etc/redhat-release ]]; then
     MTR_ARGS+=" --mysqld=--loose-mecab-rc-file=/etc/mecabrc"
 fi
 #
+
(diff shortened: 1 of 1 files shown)


local/test-binary
```diff
@@ -113,6 +113,11 @@ if [[ ! -e /etc/redhat-release ]]; then
     MTR_ARGS+=" --mysqld=--loose-mecab-rc-file=/etc/mecabrc"
 fi
 #
+
+if [[ "${MTR_ARGS}" == *"--only-big-test"* ]]; then
```
laurynas-biveinis: --only-big-test or --big-test

laurynas-biveinis (changes requested):

local/test-binary
hors: ok, fixed

laurynas-biveinis (approved):

PR #200, opened 2019-04-29
JEN-1090 use -DWITH_LIBEVENT=bundled as a build option

hors:
https://ps3.cd.percona.com/view/8.0/job/percona-server-8.0-param/51/

Or I can do it only for ubuntu:xenial, debian:stretch and for centos 6/7 OSs.


laurynas-biveinis:
@hors let's use the system one where its version is OK

EvgeniyPatlan:
@laurynas-biveinis it is hard to get libevent-2.1 version installed on all systems. So it is better to use bundled in this case.

laurynas-biveinis:
I meant use system libevent on those systems where it is >= 2.1 out of the box

hors:
https://ps3.cd.percona.com/view/8.0/job/percona-server-8.0-param/54/

EvgeniyPatlan (approved):

laurynas-biveinis (approved):

PR #204, opened 2019-05-02
JEN-1094 include CTestTestfile.cmake in binary tarball

hors:


laurynas-biveinis (commented):
How is this enough? The sample output on JEN showed an error instead of 170+ tests run

hors:
> How is this enough? The sample output on JEN showed an error instead of 170+ tests run

It is not enough and it was discussed in the task, I will re-work it.

PR #205, opened 2019-05-02
JEN-1096 remove anaconda-post.log from root dir

hors:


EvgeniyPatlan (approved):

docker/install-deps
```diff
@@ -93,6 +93,11 @@ if [ -f /usr/bin/yum ]; then
     [[ ${RHVER} -eq 8 ]] && yum -y reinstall kernel-headers
 
     yum -y clean all
+
+#   this is workaround for jen-1096 as soon as the issue is fixed we need to remove it
```
laurynas-biveinis:
```suggestion
#   this is workaround for https://bugs.mysql.com/bug.php?id=95222 as soon as the issue is fixed we need to remove it 
```

laurynas-biveinis (changes requested):

laurynas-biveinis (approved):

PR #208, opened 2019-06-03
JEN-1109 install and use devtoolset-8 for Centos 7/6

hors:
https://ps3.cd.percona.com/view/8.0/job/percona-server-8.0-param/79/

### DIFF
--- a/docker/install-deps
+++ b/docker/install-deps
@@ -81,6 +81,9 @@ if [ -f /usr/bin/yum ]; then
         PKGLIST+=" devtoolset-7-gcc-c++ devtoolset-7-binutils cmake3"
         PKGLIST+=" devtoolset-7-libasan-devel devtoolset-7-libubsan-devel"
         PKGLIST+=" devtoolset-7-valgrind devtoolset-7-valgrind-devel"
+        PKGLIST+=" devtoolset-8-gcc-c++ devtoolset-8-binutils"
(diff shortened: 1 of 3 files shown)


docker/install-deps
```diff
@@ -81,6 +81,9 @@ if [ -f /usr/bin/yum ]; then
         PKGLIST+=" devtoolset-7-gcc-c++ devtoolset-7-binutils cmake3"
         PKGLIST+=" devtoolset-7-libasan-devel devtoolset-7-libubsan-devel"
         PKGLIST+=" devtoolset-7-valgrind devtoolset-7-valgrind-devel"
```
laurynas-biveinis, line 83: devtoolset-7 is redundant then?
Also, the way it all looks structured, devtoolset-2 is always installed, regardless of it is needed or not?

laurynas-biveinis (changes requested):

docker/install-deps
hors: We add  all changes only in docker/install-deps for branch 8.0  and then use it to build docker images which are used for all branches that is why we need to install all these rpms. devtoolset-2  is used for Centos 6/i386 .

hors: We add  all changes only in docker/install-deps for branch 8.0  and then use it to build docker images which are used for all branches that is why we need to install all these rpms. devtoolset-2  is used for Centos 6/i386 .

laurynas-biveinis (approved):

PR #210, opened 2019-06-21
JEN-1117 add '-DWITH_ROUTER=ON/OFF' and '-DWITH_MYSQLX=ON/OFF' options

hors:
https://ps3.cd.percona.com/view/8.0/job/percona-server-8.0-param/96/

All trunk jobs will use default value and it is 'ON'. 

@laurynas-biveinis please ping me when you merge 8.0.16


laurynas-biveinis (commented):
What are the defaults on param page?
Unfortunately, since I wrote up the issue, I learned that some tests are not compatible with WITH_MYSQLX=OFF default, so it has to default to ON

8.0.16 is in the trunk already

hors:
> What are the defaults on param page?
> Unfortunately, since I wrote up the issue, I learned that some tests are not compatible with WITH_MYSQLX=OFF default, so it has to default to ON
> 
> 8.0.16 is in the trunk already

Thank you for information. I have changed the default value for param job to 'ON' for WITH_MYSQLX option.

laurynas-biveinis (approved):

PR #222, opened 2019-08-28
JEN-960 add '--sanitize' option in case of 'UBSAN/ASAN' builds

hors:
* add '--sanitize' option in case of '-WITH_UBSAN/-WITH_ASAN' builds
    * fix 'UBSAN' builds for rhel8

### DIFF
--- a/local/test-binary
+++ b/local/test-binary
@@ -81,7 +81,7 @@ function process_mtr_output {
   sed -i.bak -e "s:@@SKIPPED@@:${SKIPPED}:g" ${OUTFILE}
   sed -i.bak -e "s: \[::g" ${OUTFILE}
 }
-##
+
 if [[ "${ANALYZER_OPTS}" == *WITH_VALGRIND=ON* ]]; then
   MTR_ARGS+=" --valgrind --valgrind-clients --valgrind-option=--leak-check=full --valgrind-option=--show-leak-kinds=all"
   [[ ${OPENSSL_VER} < '1.0.2' ]] && export OPENSSL_ia32cap=~0x200000000000000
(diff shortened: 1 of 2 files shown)


hors:
https://ps3.cd.percona.com/job/percona-server-8.0-pipeline/2060/

george-lorch (approved):
+1 pending Yura review

hors:
https://ps3.cd.percona.com/job/percona-server-8.0-pipeline/2064 - UBSan fix for rhel8

local/test-binary
```diff
@@ -90,15 +90,19 @@ if [[ "${ANALYZER_OPTS}" == *WITH_VALGRIND=ON* ]]; then
     PARALLEL=$((PARALLEL/3))
   fi
 fi
-#
+
+if [[ "${ANALYZER_OPTS}" == *WITH_ASAN=ON* ]] || [[ "${ANALYZER_OPTS}" == *WITH_UBSAN=ON* ]]; then
```
percona-ysorokin: `${ANALYZER_OPTS}` can also be `-DWITH_MSAN=ON` (even though we do not test this currently).
I am not also sure about `WITH_ASAN_SCOPE=ON: as far as I remember `WITH_ASAN_SCOPE=ON` always comes along with `WITH_ASAN=ON` - please double check.

percona-ysorokin (changes requested):

local/test-binary
hors: you are right, WITH_ASAN_SCOPE should be used with WITH_ASAN

from documentation:
-DWITH_ASAN_SCOPE=bool
Whether to enable the AddressSanitizer -fsanitize-address-use-after-scope Clang flag for use-after-scope detection. The default is off. To use this option, -DWITH_ASAN must also be enabled.

percona-ysorokin (approved):
LGTM

PR #229, opened 2019-10-01
JEN-1147 fix CentOS 7.7 rpm dependency issue

hors:


docker/install-deps
```diff
@@ -63,13 +63,17 @@ if [ -f /usr/bin/yum ]; then
     fi
 
 # Percona-Server
+    if [[ ${RHVER} -eq 7 ]]; then
```
percona-ysorokin, line 66: What about RHEL/CentOS 8?

hors: @percona-ysorokin the issue was fixed on centos, the fix is not needed anymore. I am going to close this PR.

PR #234, opened 2019-10-21
JEN-1161 use ubuntu disco for asan tests

hors:
* for asan tests we need to use libasan5 and gcc-8+, so it is
      better to use ubuntu disco for these needs because it has
      libasan5 and gcc-8+ installed by default


hors:
https://ps3.cd.percona.com/view/5.6/job/percona-server-5.6-asan-param/1/

percona-ysorokin (approved):
LGTM

PR #272, opened 2020-09-15
PS-7267: Implement CI FS tests

Sudokamikaze:
Link to test job: https://ps80.cd.percona.com/job/percona-server-8.0-pipeline-ci/

### DIFF
--- a/docker/run-test
+++ b/docker/run-test
@@ -7,8 +7,8 @@ ROOT_DIR=$(cd $(dirname $0)/../sources; pwd -P)
 SCRIPTS_DIR=$(cd $(dirname $0)/../local; pwd -P)
 SOURCE_IMAGE=${1:-centos:7}
 
-if [[ -d /mnt/mtr_disk_dir ]] && [[ ${ENABLE_TESTS_ON_CI_FS} == "yes" ]] && [[ "${CMAKE_BUILD_TYPE}" == "RelWithDebInfo" ]]; then
-    CI_FS_DOCKER_FLAG=" --mount type=bind,source=/mnt/mtr_disk_dir,destination=/tmp/mtr_disk_dir"
+if [[ -d /mnt/mtr_disk_dir_${CMAKE_BUILD_TYPE} ]]; then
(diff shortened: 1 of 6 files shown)


docker/run-test
```diff
@@ -7,10 +7,15 @@ ROOT_DIR=$(cd $(dirname $0)/../sources; pwd -P)
 SCRIPTS_DIR=$(cd $(dirname $0)/../local; pwd -P)
 SOURCE_IMAGE=${1:-centos:7}
 
+if [[ -d /mnt/mtr_disk_dir ]] && [[ ${ENABLE_TESTS_ON_CI_FS} == "yes" ]] && [[ "${CMAKE_BUILD_TYPE}" == "RelWithDebInfo" ]]; then
```
kamil-holubicki: Does it mean that CI tests will be executed only for release builds?

jenkins/pipeline.yml
```diff
@@ -140,6 +140,16 @@
         name: MTR_REPEAT
         default: "1"
         description: Run each test N number of times, --repeat=N
+    - choice:
+        name: ENABLE_TESTS_ON_CI_FS
```
kamil-holubicki, line 144: I see it defaults to NO. I would default it to YES

kamil-holubicki (commented):
Does it work?

https://ps80.cd.percona.com/job/percona-server-8.0-pipeline-ci/129/consoleText

```
export ENABLE_TESTS_ON_CI_FS='\''yes'\''
.....
200916  9:28:49 [ 92%] main.ps_1general_ci                      w1  [ skipped ]  Test requires 'case_insensitive_file_system'

```
Additionally, if possible it would be better to use 
`grep -rl . -e include/have_case_insensitive_file_system.inc`
to find tests instead of hardcoding '_ci'

docker/run-test
Sudokamikaze: As pipeline launches, there will be two builds on the same worker, creating an interference between them

I decided to point to a label, so that only one is going, will investigate on it to find a better solution

Sudokamikaze:
Closed as it was reworked here: https://github.com/Percona-Lab/ps-build/pull/296

PR #275, opened 2020-10-23
PS-7390: Remove --non-parallel-test from default MTR args

Sudokamikaze:
If we want to have a backward compatibility(in order to run older than 8.0.22) I'd add a version check from `mysqld --version`

https://ps3.cd.percona.com/job/percona-server-8.0-pipeline/3625/ build

### DIFF
--- a/local/test-binary
+++ b/local/test-binary
@@ -56,7 +56,7 @@ if [[ -z "${JEMALLOC}" ]]; then
   exit 1
 fi
 #
-MTR_ARGS+=" --timestamp --non-parallel-test --report-unstable-tests"
+MTR_ARGS+=" --timestamp --report-unstable-tests"
 #
 ##
 function process_mtr_output {


EvgeniyPatlan (changes requested):
where is link to the jenkins?

Sudokamikaze:
> where is link to the jenkins?

Hi @EvgeniyPatlan 
I've updated dsc, here's the link https://ps3.cd.percona.com/job/percona-server-8.0-pipeline/3625

kamil-holubicki (approved):
lgtm

>If we want to have a backward compatibility(in order to run older than 8.0.22) I'd add a version check from mysqld --version

In my opinion we don't need it in Jenkinsfile (it will make pipeline configuration more complicated). This solution is still backward compatible as the flag can be set via MTR_ARGS

EvgeniyPatlan (approved):

percona-ysorokin (approved):
LGTM

PR #279, opened 2020-10-29
FB8-254: Restructure python3 installation & deps

Sudokamikaze:
CentOS 6 doesn't have python3 in BaseOS repository, only in SCLo, which is not intended for this purpose
So, this commit adds needed guards, in order to pass build stage of Dockers


hors (approved):

delgod (approved):

percona-ysorokin (approved):
LGTM

PR #296, opened 2020-12-24
PS-7267: [Reworked] Implement CI FS tests

Sudokamikaze:
Builds: https://ps80.cd.percona.com/job/percona-server-8.0-param-ci/9/

### DIFF
--- a/jenkins/pipeline.groovy
+++ b/jenkins/pipeline.groovy
@@ -110,6 +110,10 @@ pipeline {
             defaultValue: '',
             description: 'TokuDB specific mtr args',
             name: 'TOKUDB_ENGINES_MTR_ARGS')
+        choice(
+            choices: 'yes\nno',
+            description: 'Run case-insensetive MTR tests',
+            name: 'CI_FS_MTR')
         string(
             defaultValue: '--unit-tests-report',
             description: 'mysql-test-run.pl options, for options like: --big-test --only-big-test --nounit-tests --unit-tests-report',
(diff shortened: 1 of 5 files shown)


jenkins/pipeline.groovy
```diff
@@ -240,6 +244,15 @@ pipeline {
                                     sleep 5
                                 done
 
+                                if [[ \$CI_FS_MTR == 'yes' ]]; then
+                                    if [[ ! -f /mnt/ci_disk_\$CMAKE_BUILD_TYPE.img ]] && [[ -z \$(mount | grep /mnt/ci_disk_dir_\$CMAKE_BUILD_TYPE) ]]; then
+                                        sudo dd if=/dev/zero of=/mnt/ci_disk_\$CMAKE_BUILD_TYPE.img bs=1G count=10
+                                        sudo /sbin/mkfs.vfat /mnt/ci_disk_\$CMAKE_BUILD_TYPE.img
+                                        sudo mkdir -p /mnt/ci_disk_dir_\$CMAKE_BUILD_TYPE
+                                        sudo mount -o loop -o uid=27 -o gid=27 -o check=r /mnt/ci_disk_\$CMAKE_BUILD_TYPE.img /mnt/ci_disk_dir_\$CMAKE_BUILD_TYPE
```
percona-ysorokin, line 252: Magic numbers `27`
May be
```
-o uid=`id -u ...` -o gid=`id - g ...`
```

local/test-binary
```diff
@@ -195,4 +195,22 @@ if [[ "${TOKUDB_ENGINES_MTR}" = 'no' ]] && [[ "${HOTBACKUP_TESTING}" = 'no' ]] &
...
+        CI_TESTS+=" $CI_TESTS_TMP"
+    done
+
+    MTR_BUILD_THREAD=auto ./mtr \
+        --force --mysqld-env="LD_PRELOAD=${MYSQLD_ENV}" \
+        --max-test-fail=0 \
+        --parallel=$(nproc) \
+        --retry=0 \
+        --tmpdir="/tmp/ps_mtr_tempdir" \
+        --vardir="/tmp/ci_disk_dir" \
```
percona-ysorokin, line 211: Also add `--big-test`

percona-ysorokin (changes requested):

jenkins/pipeline.groovy
Sudokamikaze: Those IDs are from docker image:
```
uid=27(mysql) gid=27(mysql) groups=27(mysql)
```
If we use `id -u` it'll return id of the worker user

Sudokamikaze: And this ID/GID is hardcoded here: https://github.com/Percona-Lab/ps-build/blob/8.0/docker/Dockerfile.inc#L9

Sudokamikaze: To be exact, it'll return:
```
uid=1000(ec2-user) gid=1000(ec2-user)
```

percona-ysorokin: Actually, I meant
```
-o uid=`id -u mysql` -o gid=`id - g mysql`
```
But I think I understand that it's not possible as there is no `mysql` user outside docker image (in the worker OS).
So, hardcoded values are fine then.

local/test-binary
Sudokamikaze: Added in 3f0b1cc
Builds: https://ps80.cd.percona.com/job/percona-server-8.0-param-ci/10/

jenkins/pipeline.groovy
Sudokamikaze: It's not possible, as this user don't exist on worker node, only inside ps-build image

percona-ysorokin (approved):
LGTM

EvgeniyPatlan (approved):

PR #297, opened 2020-12-25
PS-7482: Modify prepare-ps-build-docker job

Sudokamikaze:
* Sync with deployed job on Jenkins, removal of CentOS 6

Job in this repository was slightly outdated due to manual changes which were done on Jenkins side, I've pulled them from Jenkins and removed CentOS 6 support

### DIFF
--- a/jenkins/prepare-ps-build-docker.yml
+++ b/jenkins/prepare-ps-build-docker.yml
@@ -52,16 +52,12 @@
                 stage('Build') {
                     steps {
                         parallel(
-                            "i386/centos:6":  { build('i386/centos:6') },
-                            "centos:6":       { build('centos:6') },
-                            "centos:7":       { build('centos:7') },
-                            "centos:8":       { build('centos:8') },
-                            "ubuntu:artful":  { build('ubuntu:artful') },
-                            "ubuntu:xenial":  { build('ubuntu:xenial') },
+                            "centos:7":  { build('centos:7') },
+                            "centos:8":  { build('centos:8') },
                             "ubuntu:bionic":  { build('ubuntu:bionic') },
-                            "ubuntu:focal":   { build('ubuntu:focal') },
-                            "debian:jessie":  { build('debian:jessie') },
-                            "debian:stretch": { build('debian:stretch') },
+                            "ubuntu:xenial":  { build('ubuntu:xenial') },
+                            "ubuntu:focal":  { build('ubuntu:focal') },
+                            "debian:stretch":  { build('debian:stretch') },
                             "debian:buster":  { build('debian:buster') },
                         )
                     }


jenkins/prepare-ps-build-docker.yml
```diff
@@ -35,7 +35,6 @@
                 label 'docker-32gb'
             }
             options {
-                compressBuildLog()
```
delgod: why not?

delgod (changes requested):

jenkins/prepare-ps-build-docker.yml
```diff
@@ -52,16 +51,12 @@
...
                         parallel(
-                            "i386/centos:6":  { build('i386/centos:6') },
-                            "centos:6":       { build('centos:6') },
-                            "centos:7":       { build('centos:7') },
-                            "centos:8":       { build('centos:8') },
-                            "ubuntu:artful":  { build('ubuntu:artful') },
-                            "ubuntu:xenial":  { build('ubuntu:xenial') },
+                            "centos7":  { build('centos:7') },
+                            "centos8":  { build('centos:8') },
                             "ubuntu:bionic":  { build('ubuntu:bionic') },
```
delgod, line 57: can we use numbers here?
like `ubuntu:18.04` instead of `ubuntu:bionic`
reason - it is simpler to compare with page - https://www.percona.com/services/policies/percona-software-support-lifecycle

delgod (changes requested):

jenkins/prepare-ps-build-docker.yml
Sudokamikaze: We can, but it will require to rework all pipelines in ps80 jenkins which use this docker

jenkins/prepare-ps-build-docker.yml
Sudokamikaze: This came from sync with jenkins, will check & re-add it back

Sudokamikaze: Re-added in 9fb97b4

jenkins/prepare-ps-build-docker.yml
```diff
@@ -52,16 +52,12 @@
...
                     steps {
                         parallel(
-                            "i386/centos:6":  { build('i386/centos:6') },
-                            "centos:6":       { build('centos:6') },
-                            "centos:7":       { build('centos:7') },
-                            "centos:8":       { build('centos:8') },
-                            "ubuntu:artful":  { build('ubuntu:artful') },
-                            "ubuntu:xenial":  { build('ubuntu:xenial') },
+                            "centos7":  { build('centos:7') },
+                            "centos8":  { build('centos:8') },
```
delgod: looks like missed colon
```suggestion
                            "centos:7":  { build('centos:7') },
                            "centos:8":  { build('centos:8') },
```

delgod (changes requested):

jenkins/prepare-ps-build-docker.yml
Sudokamikaze: @delgod done

EvgeniyPatlan (approved):

delgod (approved):

PR #321, opened 2021-04-10
ENG-741: Implement feature of getting failed/skipped tests names in separate file

Sudokamikaze:
Param job URL:https://ps80.cd.percona.com/job/percona-server-8.0-param-ENG-741/5/

### DIFF
--- a/jenkins/param.yml
+++ b/jenkins/param.yml
@@ -175,6 +175,7 @@
         do-not-fingerprint: true
     - shell: |
         echo "${TRIGGERED_BUILD_NUMBERS_percona_server_8_0_pipeline}" > PIPELINE_BUILD_NUMBER
+        find . -name "*.txt" -exec echo {} \;
         gunzip build.log.gz
     publishers:
     - raw:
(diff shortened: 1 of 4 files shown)


jenkins/param.yml
```diff
@@ -205,3 +206,5 @@
         keep-long-stdio: true
     - archive:
         artifacts: 'PIPELINE_BUILD_NUMBER'
+    - archive:
+        artifacts: "**/*.txt"
```
percona-ysorokin: No EOL

percona-ysorokin (approved):
LGTM with one minor comment resolved

Sudokamikaze:
Discussed with Przemyslaw, this solution doesn't meet expectations and it's not possible to crawl all logs into single file and then sort it, then archive inside whole job, not for each configuration respectively

We came to conclusion, that this feature doesn't meet expectations and faster way will be bash script which gathers logs from `#showfailedlinks` and parses it

Closing as we agreed

PR #322, opened 2021-04-12
ENG-1002: [PS80] Migrate to warnings-ng plugin

Sudokamikaze:
URL: https://ps3.cd.percona.com/view/8.0/job/percona-server-8.0-param/200/

### DIFF
--- a/jenkins/asan-param.yml
+++ b/jenkins/asan-param.yml
@@ -164,12 +164,29 @@
         echo "${TRIGGERED_BUILD_NUMBERS_percona_server_8_0_pipeline}" > PIPELINE_BUILD_NUMBER
         gunzip build.log.gz
     publishers:
-    - warnings:
-        workspace-file-scanners:
-        - file-pattern: 'build.log'
-          scanner: 'GNU C Compiler 4 (gcc)'
-        run-always: true
-        shouldDetectModules: false
+    - raw:
+        xml: !!binary |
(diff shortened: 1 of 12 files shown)


jenkins/asan-param.yml
```diff
@@ -164,12 +164,29 @@
         echo "${TRIGGERED_BUILD_NUMBERS_percona_server_8_0_pipeline}" > PIPELINE_BUILD_NUMBER
         gunzip build.log.gz
     publishers:
-    - warnings:
-        workspace-file-scanners:
-        - file-pattern: 'build.log'
-          scanner: 'GNU C Compiler 4 (gcc)'
-        run-always: true
-        shouldDetectModules: false
+    - raw:
```
delgod, line 167: can we use `recordIssues enabledForFailure: true, tools: [gcc(pattern: 'build.log')]` here instead of raw xml?

delgod (changes requested):

jenkins/asan-param.yml
Sudokamikaze: We can't, I tried already

Sudokamikaze: Sources does not contain following syntax: https://opendev.org/jjb/jenkins-job-builder/src/branch/master/jenkins_jobs/modules/publishers.py

delgod (approved):

EvgeniyPatlan (approved):

PR #327, opened 2021-04-14
PS-7639: Build libeatmydata 128 version for now

Sudokamikaze:
129 version, which was pushed 4 days ago fails to build on CentOS 6 due to old autoconf

### DIFF
--- a/docker/install-deps
+++ b/docker/install-deps
@@ -223,6 +223,7 @@ fi
 if [ ! -f /usr/local/lib/libeatmydata.so ]; then
     git clone https://github.com/stewartsmith/libeatmydata /tmp/libeatmydata
     pushd /tmp/libeatmydata
+        # PS-7639: Needed for CentOS 6 because of ancient autoconf
         if [ -f /usr/bin/yum ]; then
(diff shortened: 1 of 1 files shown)


docker/install-deps
```diff
@@ -223,6 +223,7 @@ fi
 if [ ! -f /usr/local/lib/libeatmydata.so ]; then
     git clone https://github.com/stewartsmith/libeatmydata /tmp/libeatmydata
     pushd /tmp/libeatmydata
+        git checkout df7ddeb0345104f25b5b7bb154bc6a008c8c8404
```
EvgeniyPatlan: please add comment why do we need exactly this commit

Sudokamikaze: Screenshot 2021-04-14 at 20 53 14
This is the last working commit on CentOS 6

EvgeniyPatlan (approved):

Sudokamikaze:
We decided to build 128 version only for CentOS 6, other platforms will be using fresh version because of performance improvements

percona-ysorokin (approved):
LGTM

PR #336, opened 2021-05-19
PS-7718: [5.7] Correct conditional check for cleaning up docker images on docker workers

Sudokamikaze:
This change is needed, when vault|consul|ps-build image was spawned, but stuck in "Created" state
`docker ps -q` will return 0, even if there's a container

This can happen on dirty worker, where PS pipeline was aborted, on bootstrap of vault
Next run will fail, due to existing containers with same name

Before:
```
sudokamikaze at Sudokamikazes-Work-MacBook-Pro in /tmp
$ docker ps -q | wc -l
       0

sudokamikaze at Sudokamikazes-Work-MacBook-Pro in /tmp
$ docker ps -a
CONTAINER ID   IMAGE         COMMAND                  CREATED         STATUS    PORTS     NAMES
49ad8ff4ce7f   vault:0.9.6   "docker-entrypoint.s…"   2 minutes ago   Created             vault-dev-v1
```

After:
```
$ docker ps -a -q | wc -l
       1
```

No builds are needed(from my point of view) as it can be reproduced locally

### DIFF
--- a/jenkins/pipeline.groovy
+++ b/jenkins/pipeline.groovy
@@ -267,8 +267,8 @@ pipeline {
                                 echo Test: \$(date -u "+%s")
                                 aws ecr-public get-login-password --region us-east-1 | docker login -u AWS --password-stdin public.ecr.aws/e7j3v3n0
                                 sg docker -c "
-                                    if [ \$(docker ps -q | wc -l) -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


percona-ysorokin (approved):
LGTM

EvgeniyPatlan (approved):

PR #343, opened 2021-08-30
PS-7876: Correct permissions of built artifacts

Sudokamikaze:
This was introduced in https://github.com/Percona-Lab/ps-build/commit/e3414b3101fc431c44096b4f01a68c8bd44a3928
And affects current build pipelines

I'm reverting that particular change to unblock percona-server-8.0-param
ZenFS enabled build should be working fine, I'll test it after merge of this change as it's a critical one


EvgeniyPatlan (approved):

percona-ysorokin (approved):
LGTM

PR #352, opened 2021-11-01
PS-7959: Use bundled lz4 lib for PS builds

oleksandr-kachan:
https://jira.percona.com/browse/PS-7959

The minimum required lz4 lib version increased to 1.9.3.
Use bundled version for builds as most of supported platforms
will fail this requirement.


percona-ysorokin (approved):
LGTM

EvgeniyPatlan (approved):

PR #359, opened 2022-03-22
FB8-277 - Create symbolic link /tmp/results/unittest/gunit

ldonoso:
https://jira.percona.com/browse/FB8-277

Some tests are failing because the filesystem is set erroneously as
lowercase becase that directory doesn't exists.

The test FW tries to create a file on that dir and it expects the dir
exists.

Please, mind that we are copying the contents of that directory in the
build script:

```
cp -r ${WORKDIR}/unittest/gunit/*  ${INSTALL_DIR}/usr/local/${PRODUCT_FULL}/
```


EvgeniyPatlan (approved):

inikep (approved):
LGTM

PR #366, opened 2022-07-06
ENG-1135 - Remove CentOS 6 from supported OS for PS 8.0 Jenkins jobs

ldonoso:
The reasons for removing it are:

- CentOS 6 is not officially supported.

- The cmake version of CentOS 6 is not supported by 8.0.28-20:

```
CMake Error at extra/opensslpp/CMakeLists.txt:17 (cmake_minimum_required):
  CMake 3.8 or higher is required.  You are running version 3.6.1
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,8 @@ git clone https://github.com/Percona-Lab/ps-build
 cd ps-build
 ./local/checkout
 
-./docker/run-build centos:6
-./docker/run-test centos:6
+./docker/run-build centos:7
+./docker/run-test centos:7
 ```
 
 ## Docker debug
@@ -14,7 +14,7 @@ git clone https://github.com/Percona-Lab/ps-build
 cd ps-build
 ./local/checkout
 
-./docker/run-build-debug centos:6
+./docker/run-build-debug centos:7
 ```
 
 # Local way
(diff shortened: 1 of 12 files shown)


percona-ysorokin (commented):
@ldonoso We also have a lot of centos:6 logic inside `run-test` / `run-build` / `install-deps` scripts. Shouldn't we remove it as well?

EvgeniyPatlan (approved):

EvgeniyPatlan:
> @ldonoso We also have a lot of centos:6 logic inside `run-test` / `run-build` / `install-deps` scripts. Shouldn't we remove it as well?

@percona-ysorokin for now it is ok to leave

ldonoso:
> > @ldonoso We also have a lot of centos:6 logic inside `run-test` / `run-build` / `install-deps` scripts. Shouldn't we remove it as well?
> 
> @percona-ysorokin for now it is ok to leave

ldonoso:
> @ldonoso We also have a lot of centos:6 logic inside `run-test` / `run-build` / `install-deps` scripts. Shouldn't we remove it as well?

@percona-ysorokin , I remember you asked that to @georgelorchpercona during a standup. We agreed to just remove Centos 6 from the jenkins interface but keep all the necessary dependencies, custom configuration, etc just in case we need to restore the OS it in the future.

percona-ysorokin (approved):
LGTM

PR #383, opened 2023-05-30
PXC-4213: Make possible to disable tests executed in ci_fs

kamil-holubicki:
https://jira.percona.com/browse/PXC-4213

Problem:
If CI_FS test is disabled in disabled.def, it is executed anyway

Cause:
We search for CI_FS tests and explicitly specify the list to be executed by mtr framework. This cause disabled.def exclusions to be skipped.

Solution:
Filter out tests disabled by disabled.def file

### DIFF
--- a/local/test-binary-parallel-mtr
+++ b/local/test-binary-parallel-mtr
@@ -288,14 +288,23 @@ if [[ "${CI_FS_MTR}" = 'yes' ]]; then
         MTR_ARGS_CI+=" --big-test"
     fi
 
-    CI_TESTS=""
-    for CI_TESTS_TMP in $(grep --exclude="*.inc" --exclude="*.log" -rl . -e include/have_case_insensitive_file_system.inc | awk -F '/' '{print $NF}' | sed 's/.test//g'); do
-        CI_TESTS+=" $CI_TESTS_TMP"
-    done
+    # Collect all CI_FS tests
+    CI_TESTS=$(grep --exclude="*.inc" --exclude="*.log" -rl . -e include/have_case_insensitive_file_system.inc | awk -F '/' '{print $(NF-2)"."$NF}' | sed 's/\.test//g' | sed 's/^\./main/g')
     # Requested in PS-7602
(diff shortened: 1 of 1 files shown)


kamil-holubicki:
This was discovered in PXC pipeline and reported in https://jira.percona.com/browse/PXC-4213, 
But the same problem was in PS pipeline.
That's why the branch name mentions PXC.

EvgeniyPatlan (approved):

local/test-binary-parallel-mtr
```diff
@@ -288,14 +288,21 @@ if [[ "${CI_FS_MTR}" = 'yes' ]]; then
...
-    for CI_TESTS_TMP in $(grep --exclude="*.inc" --exclude="*.log" -rl . -e include/have_case_insensitive_file_system.inc | awk -F '/' '{print $NF}' | sed 's/.test//g'); do
-        CI_TESTS+=" $CI_TESTS_TMP"
-    done
+    # Collect all CI_FS tests
+    CI_TESTS=$(grep --exclude="*.inc" --exclude="*.log" -rl . -e include/have_case_insensitive_file_system.inc | awk -F '/' '{print $(NF-2)"."$NF}' | sed 's/\.test//g' | sed 's/^\./main/g')
     # Requested in PS-7602
-    if [[ ${CMAKE_BUILD_TYPE} = 'Debug' ]]; then
-        CI_TESTS+=" information_schema.i_s_schema_definition_debug"
-    fi
+    CI_TESTS+=" information_schema.i_s_schema_definition_debug"
```
percona-ysorokin: Any reason why we now add this `information_schema.i_s_schema_definition_debug` to the RelWithDebInfo as well?

kamil-holubicki: Actually, I dropped `if` by omission. It makes no difference as MTR test has source `include/have_debug.inc`, so it will be skipped by MTR framework anyway.
Changed back to `if debug` version.

percona-ysorokin (approved):
LGTM

PR #388, opened 2023-09-29
PS-7 ps57.cd: use hashicorp/consul instead of consul

oleksandr-kachan:
According to https://hub.docker.com/_/consul:
Starting with Consul 1.16, we will stop publishing official Dockerhub images and publish only our Verified Publisher images. Users of Docker images should pull from hashicorp/consul instead of consul.

(cherry picked from commit 10305daace235fcadb58f77419cd7811659ee6bb)

### DIFF
--- a/local/bootstrap-vault
+++ b/local/bootstrap-vault
@@ -40,7 +40,7 @@ parse_arguments() {
 
 function prepare_consul {
     docker rm -f consul ||:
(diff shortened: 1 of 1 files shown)


percona-ysorokin (approved):
LGTM

vorsel (approved):

PR #394, opened 2023-11-06
PS-8978: Fix router unit tests with improved `-DCOMPILATION_COMMENT`

inikep:
We set `-DCOMPILATION_COMMENT=Percona Server (GPL), Release XXXX, Revision XXXXXX-debug` what breaks router unit tests because `(` and `)` are special characters in regex:
```
Expected: contains at least one element that contains regular expression "# Bootstrapping MySQL Router 8.0.35 \\(Percona Server (GPL), Release , Revision XXXXXX-debug\\) instance at"
  Actual: { "Please enter MySQL password for root: ", "WARNING: The MySQL server does not have SSL configured and metadata used by the router may be transmitted unencrypted.", "", "***** WARNING *****", "Account verification failed with error:", "  Error connecting to MySQL server at 127.0.0.1:11039: Access Denied for user 'mysql_router1_93uy5zylnbjx'@'localhost' (1045)", "", "This means that we were unable to log in using the accounts that were created", "and run SQL queries that Router needs to run during its operation.", "It means this Router instance may be inoperable and user intervention is", "required to correct the issue and/or bootstrap again.", "", "See https://dev.mysql.com/doc/mysql-router/8.0/en/ for more information.", "# Bootstrapping MySQL Router 8.0.35 (Percona Server (GPL), Release , Revision XXXXXX-debug) instance at '/tmp/router-1IhzFo'...", "", "- Creating account(s) (only those that are needed, if any)", "- Verifying account (using it to run SQL queries that would be run by Router)", "-
```


EvgeniyPatlan (approved):

percona-ysorokin (approved):
LGTM

PR #395, opened 2023-11-07
PS-8978: Add option to run MySQL Router unit tests

inikep:
Add `UNIT_TESTS_ROUTER` option (off by default) to decide whether MySQL Router unit tests should be run

### DIFF
--- a/local/build-binary
+++ b/local/build-binary
@@ -235,8 +235,10 @@ if [ -d ${SOURCEDIR}/rqg ]; then
 fi
 # Copy unit tests
 cp ${WORKDIR}/CTestTestfile.cmake ${INSTALL_DIR}/usr/local/${PRODUCT_FULL}/
-cp -r ${WORKDIR}/router ${INSTALL_DIR}/usr/local/${PRODUCT_FULL}/
 cp -r ${WORKDIR}/unittest ${INSTALL_DIR}/usr/local/${PRODUCT_FULL}/
+if [[ "${UNIT_TESTS_ROUTER}" = "yes" ]]; then
+    cp -r ${WORKDIR}/router ${INSTALL_DIR}/usr/local/${PRODUCT_FULL}/
+fi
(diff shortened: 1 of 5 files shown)


local/build-binary
```diff
@@ -235,8 +235,10 @@ if [ -d ${SOURCEDIR}/rqg ]; then
 fi
 # Copy unit tests
 cp ${WORKDIR}/CTestTestfile.cmake ${INSTALL_DIR}/usr/local/${PRODUCT_FULL}/
-cp -r ${WORKDIR}/router ${INSTALL_DIR}/usr/local/${PRODUCT_FULL}/
 cp -r ${WORKDIR}/unittest ${INSTALL_DIR}/usr/local/${PRODUCT_FULL}/
+if [[ "${UNIT_TESTS_ROUTER}" = "yes" ]]; then
```
kamil-holubicki, line 239: I see that flag `UNIT_TESTS_ROUTER` only affects this place - it causes copying of router directory. I understand that if the directory is present, unit tests will be picked up automatically and executed if `MTR_ARGS` contain `--unit-tests-report `.
How long does it take to do router unit tests? Maybe it is more consistent to run them always?

Consider the following scenario:
1. Build N with `UNIT_TESTS_ROUTER=OFF`
2. Build N+1, which reuses binaries from build N and `UNIT_TESTS_ROUTER=ON`. How would it work? I would expect router UT to be run, but as I understand they will be not.

inikep: Yes, unit tests will be picked up automatically because of `CTestTestfile.cmake` which contains all possible unit tests. In cases of absence of `router` files it will be skipped.

> How long does it take to do router unit tests? Maybe it is more consistent to run them always?

It took about 2 hours on Jenkins for both Debug and RelWithDebInfo. I will create a temporary pipelines to test if unit tests are faster with more memory and `parallel=auto`.

> Consider the following scenario:

It will not work but I think we don't want to include additional 1.6+ GB of storage (100 MB after tar.gz) just in case.

kamil-holubicki (approved):

percona-ysorokin (approved):
LGTM

PR #397, opened 2023-11-17
PS-8978: Improve `percona-server-8.0-pipeline-parallel-mtr`

inikep:
1. Fix missing `install` in `sudo yum -y dosfstools`
2. Set `--shm-size=32G`
3. Limit parallel to 10 threads instead of 8 (it reduces total run time from 11h to 10h)
4. Update `suites-groups.sh`
5. Test all normal suites together and all big suites together instead testing one by one (it reduces total run time from 10h to 9h)

### DIFF
--- a/docker/run-test-parallel-mtr
+++ b/docker/run-test-parallel-mtr
@@ -55,8 +55,8 @@ if [[ "${KEYRING_VAULT_MTR}" == 'yes' ]]; then
     VAULT_DOCKER_FLAG="--network bridge-vault"
 fi
 
-docker run --rm --shm-size=24G \
-    --tmpfs /dev/shm:rw,exec,size=24G \
+docker run --rm --shm-size=32G \
+    --tmpfs /dev/shm:rw,exec,size=32G \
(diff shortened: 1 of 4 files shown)


kamil-holubicki (commented):
I've got only one concern about this:
```Test all normal suites together and all big suites together instead testing one by one (it reduces total run time from 10h to 9h)```
For sure, it is better, 10% faster. No questions.
But this way we lose the ability to measure how long particular suite is executed. And this info is need to be able to split suites among workers (or adjust this split)
Maybe it would be possible to write this script in a way that relatively simply we could execute them one by one? It may be hardcoded "switch" for this behavior, but I think there may be a time in the future when we want to measure execution times again.

kamil-holubicki (approved):

percona-ysorokin (approved):
LGTM

inikep:
Everything seems to be OK after merging:
https://ps80.cd.percona.com/view/8.0%20parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/360/

PR #398, opened 2023-11-20
PS-8978: Enable testing of router unit tests in Jenkins

inikep:
1. Copy missing `router/tests/component/data` files to tar.gz archive.
2. Router unit tests have hard coded paths to SOURCEDIR. We link proper paths in case SOURCEDIR is not available (e.g. Jenkins docker builds).
3. Remove `UNIT_TESTS_ROUTER` parameter from pipelines.

### DIFF
--- a/local/test-binary
+++ b/local/test-binary
@@ -16,9 +16,23 @@ set -o errexit
 set -o xtrace
 
 WORKDIR_ABS=$(cd ${1:-./build}; pwd -P)
+SOURCEDIR=$2
 rm -fr ${WORKDIR_ABS}/PS
 mkdir -p ${WORKDIR_ABS}/PS/sql
 tar -C ${WORKDIR_ABS}/PS --strip-components=1 -zxpf $(ls $WORKDIR_ABS/*.tar.gz | head -1)
+
+if [[ "${SOURCEDIR}" != "" ]]; then
+    # Router unit tests have hard coded paths to SOURCEDIR.
+    # We link proper paths in case SOURCEDIR is not available (e.g. Jenkins docker builds).
+    if ! [ -d ${SOURCEDIR}/router/tests/component/data ]; then
+        sudo mkdir -p ${SOURCEDIR}/router/tests/component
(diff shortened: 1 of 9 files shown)


percona-ysorokin (approved):
LGTM

local/test-binary
```diff
@@ -16,10 +16,24 @@ set -o errexit
 set -o xtrace
 
 WORKDIR_ABS=$(cd ${1:-./build}; pwd -P)
+SOURCEDIR=$2
 rm -fr ${WORKDIR_ABS}/PS
 mkdir -p ${WORKDIR_ABS}/PS/sql
 tar -C ${WORKDIR_ABS}/PS --strip-components=1 -zxpf $(ls $WORKDIR_ABS/*.tar.gz | head -1)
 
+if [[ "${SOURCEDIR}" != "" ]]; then
```
kamil-holubicki, line 24: ``if [[ -n "${SOURCEDIR}" ]]``

local/test-binary-parallel-mtr
```diff
@@ -25,10 +25,24 @@ ulimit -a
 echo "*******************************************************"
 
 WORKDIR_ABS=$(cd ${1:-./build}; pwd -P)
+SOURCEDIR=$2
 rm -fr ${WORKDIR_ABS}/PS
 mkdir -p ${WORKDIR_ABS}/PS/sql
 tar -C ${WORKDIR_ABS}/PS --strip-components=1 -zxpf $(ls $WORKDIR_ABS/*.tar.gz | head -1)
 
+if [[ "${SOURCEDIR}" != "" ]]; then
```
kamil-holubicki, line 33: likewise

kamil-holubicki (approved):
Looks good.
Only one minor comment.

local/test-binary
inikep: Why do you prefer `-n`?
I prefer `[[ STRING != STRING ]]` because of readability.

kamil-holubicki: This is just a minor comment.
The Bash way of checking if variable is not empty is `-n`.
Both ways work fine :)

PR #399, opened 2023-11-21
PS-8978: Use different `WORKER_x_MTR_SUITES` for `Debug` and `RelWithDebInfo`

inikep:
1. Set different `WORKER_x_MTR_SUITES` for different `CMAKE_BUILD_TYPE`.
2. Improve logging for `setupTestSuitesSplit()`.

### DIFF
--- a/jenkins/pipeline-parallel-mtr.groovy
+++ b/jenkins/pipeline-parallel-mtr.groovy
@@ -219,7 +219,7 @@ void build(String SCRIPT) {
 }
 
 void setupTestSuitesSplit() {
-    sh """#!/bin/bash
+    def split_script = """#!/bin/bash
         if [[ "${FULL_MTR}" == "yes" ]]; then
             # Try to get suites split from PS repo. If not present, fallback to hardcoded.
             RAW_VERSION_LINK=\$(echo \${GIT_REPO%.git} | sed -e "s:github.com:raw.githubusercontent.com:g")
@@ -238,7 +238,8 @@ void setupTestSuitesSplit() {
             wget \${RAW_VERSION_LINK}/${BRANCH}/mysql-test/mysql-test-run.pl -O ${WORKSPACE}/mysql-test-run.pl
             chmod +x ${WORKSPACE}/suites-groups.sh
             set +e
-            ${WORKSPACE}/suites-groups.sh check ${WORKSPACE}/mysql-test-run.pl
+            echo "Check if suites list is consistent with the one specified in mysql-test-run.pl"
+            ${WORKSPACE}/suites-groups.sh check ${WORKSPACE}/mysql-test-run.pl ${CMAKE_BUILD_TYPE}
             CHECK_RESULT=\$?
             set -e
             echo "CHECK_RESULT: \${CHECK_RESULT}"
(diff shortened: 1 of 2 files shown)


jenkins/suites-groups.sh
```diff
@@ -132,11 +149,5 @@ check() {
   fi
   echo "************************"
 
-  exit ${failure}
+  return ${failure}
 }
-
-case "$1" in
```
kamil-holubicki, line 138: I would leave this part. It allows us to run the script locally just to check if everything is OK without the need of sourcing it to the current shell.

jenkins/pipeline-parallel-mtr.groovy
```diff
@@ -238,7 +238,9 @@ void setupTestSuitesSplit() {
             wget \${RAW_VERSION_LINK}/${BRANCH}/mysql-test/mysql-test-run.pl -O ${WORKSPACE}/mysql-test-run.pl
             chmod +x ${WORKSPACE}/suites-groups.sh
             set +e
-            ${WORKSPACE}/suites-groups.sh check ${WORKSPACE}/mysql-test-run.pl
+            # Set WORKER_x_MTR_SUITES and check if suites list is consistent with the one specified in mysql-test-run.pl
+            source ${WORKSPACE}/suites-groups.sh
```
kamil-holubicki: Yes, we source it to get split variables, but I'm not sure if it is the right way to call the function directly from sourced script. In my opinion more readable is calling a script as it was done before

jenkins/pipeline-parallel-mtr.groovy
```diff
@@ -238,7 +238,9 @@ void setupTestSuitesSplit() {
...
+            check_suites ${CMAKE_BUILD_TYPE} ${WORKSPACE}/mysql-test-run.pl
```
kamil-holubicki: Check should not depend on build type, right? We always use the same set of suites.

jenkins/suites-groups.sh
inikep: I've never seen such type of a call:
```
${WORKSPACE}/suites-groups.sh check ${WORKSPACE}/mysql-test-run.pl
```
The traditional usage is to `source` functions from other scripts/includes and not to `source` variables like `WORKER_x_MTR_SUITES`.
I converted `suites-groups.sh` just to consist of 2 functions `get_suites()` and `check_suites()`. The `WORKER_x_MTR_SUITES` variables will be defined only if one of functions was called.

jenkins/pipeline-parallel-mtr.groovy
inikep: We source only functions now, not variables. 
https://stackoverflow.com/questions/10822790/can-i-call-a-function-of-a-shell-script-from-another-shell-script

jenkins/pipeline-parallel-mtr.groovy
inikep: Did you read the topic of this PR?
```
PS-8978: Use different WORKER_x_MTR_SUITES for Debug and RelWithDebInfo
```

inikep: `check_suites` calls `get_suites` to define `WORKER_x_MTR_SUITES` so it needs `${CMAKE_BUILD_TYPE}`,

inikep:
I updated `suites-groups.sh` to use a hybrid approach. The following cases will work identically:
1st use case:
```
suites-groups.sh check ${WORKSPACE}/mysql-test-run.pl ${CMAKE_BUILD_TYPE}
source ${WORKSPACE}/suites-groups.sh
set_suites ${CMAKE_BUILD_TYPE}
```
2nd use case (Kamil's idea with `BUILD_TYPE` bash env variable):
```
export BUILD_TYPE=${CMAKE_BUILD_TYPE}
suites-groups.sh check ${WORKSPACE}/mysql-test-run.pl
source ${WORKSPACE}/suites-groups.sh
```
3rd use case (my preferred approach):
```
source ${WORKSPACE}/suites-groups.sh
set_suites ${CMAKE_BUILD_TYPE}
check_suites ${WORKSPACE}/mysql-test-run.pl
```

kamil-holubicki (approved):

percona-ysorokin (approved):
LGTM

PR #401, opened 2023-11-24
PS-8978: Execute unit tests seperately and before MTR tests to reduce  `/dev/shm` memory usage

inikep:
Every thread for MTR testing needs about 1 GB therefore we execute all unit tests at the beginning with `main.1st` only to reduce memory usage with `/dev/shm` and lower probability of error because of OOM.


kamil-holubicki (approved):

percona-ysorokin (approved):
LGTM

PR #408, opened 2023-12-04
PS-8978: Improve percona-server-8.x-pipeline-parallel-mtr

inikep:
Port recent improvements in `percona-server-8.0-pipeline-parallel-mtr`:
1. Every thread for MTR testing needs about 1 GB therefore we execute all unit tests at the beginning with `main.1st` only to reduce memory usage with `/dev/shm` and lower probability of error because of OOM.
2. Set different `WORKER_x_MTR_SUITES` for different `CMAKE_BUILD_TYPE`.
3. Add missing router unit tests. 
4. Limit parallel to 10 threads, use `--shm-size=32G`.
5. Do not test suites one by one. Test all normal suites together and all big suites together.
6. Fix router unit tests with improved `-DCOMPILATION_COMMENT`
7. Fix `sudo yum -y dosfstools`


percona-ysorokin (approved):
LGTM

oleksandr-kachan (approved):

PR #410, opened 2023-12-13
PS-9030 [8.0]: Keep build files outside a Docker container and execute unit tests using an original build

inikep:
Changes:
1. Reuse AWS instance for Jenkins `Build` stage and `Test 1` stage.
2. Keep build files outside a Docker container to share a build directory with `Test 1` stage.
3. Use an original build (not files from `.tar.gz` archive) to execute unit tests in `Test 1` stage.
4. Don't copy unit tests to `.tar.gz` archive (what makes the file much smaller (e.g. 1.7 GB -> 0.7 GB)).
5. Introduce a `${WORK_DIR}` workspace to keep a build, an archive, results is a single directory.
6. Remove `docker/run-test` and `local/test-binary` as they are not compatible with these changes. 
7. Improve `docker/run-build-debug`.

Usage:
```
./docker/run-build [Docker-OS] [Work-Dir]
./docker/run-test-parallel-mtr [Docker-OS] [Worker-Number] [Work-Dir]
```
When `Work-Dir` is not specified, the default value will be `${ROOT_DIR}/${SOURCE_IMAGE}_${CMAKE_BUILD_TYPE}` e.g. `./ubuntu-jammy_RelWithDebInfo`.

The `$KEEP_BUILD` option has been added, which used with `KEEP_BUILD=yes` prevents files from `${WORK_DIR}/build` from being deleted, so we can create a build once and use it multiple times for testing (also useful locally without Jenkins).
We may also keep many work dirs for different operating systems and build types.

Unit tests are executed at the beginning of `Test 1` stage if the build directory in `${WORK_DIR}/build` is available.
Now we can't execute unit tests without `Build` stage with prebuild binaries, but all 277 tests will be run.
After unit tests are finished `${WORK_DIR}/build` will be removed unless `KEEP_BUILD=yes` is used.
Then everything on `WORKER_1`/`Test 1` works the same like on other `WORKER_x`, i.e. `*.tar.gz` is decompressed and MTR tests are run using extracted binaries.

For `percona-server-8.0-pipeline-parallel-mtr` we set `KEEP_BUILD=yes` for `Build` stage and `KEEP_BUILD=no` (it's a default value) for `Test 1` stage because AWS instance is finished after `Test 1` stage and we lose the build directory anyway.


inikep:
Tested at https://ps80.cd.percona.com/job/percona-server-8.0-pipeline-testing-inikep/42/

kamil-holubicki (approved):
LGTM.
Please make commit message more clear here:
`Build server in AWS instance (not in Docker) to share a build directory with Test 1 stage.`
Build still is done in Docker. The difference is that previously the output directory was inside the container, so it was deleted when the container was removed. Now we bind the output directory to the container, so it remains after container termination and we can bind it again to the container which does tests.

percona-ysorokin (approved):
LGTM

PR #414, opened 2024-02-26
ENG-1225 update pipeline for 57eol amzn private repo

vorsel:


local/test-binary
VarunNagaraju, line 1: Can we add a check for the EOL branch in test-binary script as well and decide whether to pass junit-* options?

EvgeniyPatlan (approved):

local/test-binary
vorsel: sure, almost done

VarunNagaraju: @vorsel Perhaps, it is better if the check verified whether the branch is `if [[ "${GIT_REPO}" =~ "post-eol" ]];` since it seems like regular PS testruns on amazon linux 2 platform can make use of the junit options to MTR?

vorsel: @VarunNagaraju 
Seems you are right.
Updated.

VarunNagaraju: Great, thanks!

VarunNagaraju (approved):
LGTM.

PR #422, opened 2024-06-24
[ENG-1240]: Changing if statement to make it run on systems where bas…

maqeel75:
…h is not default shell.


vorsel (approved):

VarunNagaraju (approved):
There are multiple instances of the same check in the following file. Perhaps, it is worth checking and updating them as well. 
- jenkins/pipeline.groovy
- local/checkout
- local/test-binary

PR #426, opened 2024-07-01
PS-9218: Enable telemetry for Jenkins builds

oleksandr-kachan:
https://perconadev.atlassian.net/browse/PS-9218


vorsel (approved):

percona-ysorokin (approved):
LGTM

Should / did we do the same in 8.0?

kamil-holubicki (commented):
This will cause telemetry to be always built-in. If this is the intention, I'm fine :)

PR #428, opened 2024-09-10
PS-9328: Merge 8.4.2 - Ensure that PATH env variable is set when runn…

dlenev:
…ing unit tests.

Do not empty environment before executing ./mysql-test-run.pl for unit tests.

This should fix routertest_router_default_paths unit test failures, as this test relies on PATH environment variable being set.


inikep (approved):
LGTM

percona-ysorokin (approved):
LGTM

PR #429, opened 2024-09-10
PS-9293: 8.0.39 Merge - Ensure that PATH env variable is set when running unit tests (8.0 version)

dlenev:
Do not empty environment before executing ./mysql-test-run.pl for unit tests.

This should fix routertest_router_default_paths unit test failures, as this test relies on PATH environment variable being set.


inikep (approved):
LGTM

percona-ysorokin (approved):
LGTM

PR #449, opened 2025-02-19
PS-9500: Remove support for centos:7, ubuntu:bionic and debian:buster 

inikep:
Clean Jenkins pipelines.


percona-ysorokin (approved):
LGTM

kamil-holubicki (commented):
Looks fine, but what is this? I thought we use only parallel pipelines and these are obolete.

inikep:
I think deciding which pipelines should be removed is a separate issue.
But agree `trunk-*` pipelines are not used or disabled now.

vorsel (approved):

PR #453, opened 2025-05-12
PS-9598 [8.4]: Remove BoringSSL, `prepare-docker` and `install-deps` 

inikep:
1. Remove `prepare-docker` and `install-deps` as 8.x uses images shared with 8.0
2. Remove support for BoringSSL


percona-ysorokin (approved):
KGTM

PR #454, opened 2025-05-12
PS-9598 [8.4]: Add support for both PS 8.4 and 9.x in `suites-groups.sh`

inikep:
1. Convert `JENKINS_SCRIPTS_REPO` and `JENKINS_SCRIPTS_BRANCH` to params.
2. Add support for both PS 8.4 and 9.x in `suites-groups.sh`.


percona-ysorokin (approved):
LGTM

PR #455, opened 2025-05-14
PS-9598 [8.x]: Convert `pipeline-parallel-mtr` and `param-parallel-mtr` to JJB template for 8.4 and 9.x

inikep:


percona-ysorokin (approved):
LGTM

PR #457, opened 2025-05-14
PS-9692 [8.x]: Add `percona-server-sanitizers-epyc9654-noble.yml` pipeline

inikep:
1. Add a new pipeline that uses `as-1015cs-tnr` node (Epyc 9654) for Valgrind and ASan. The pipeline doesn't use Docker but runs directly on Ubuntu Noble.
2. The whole build and test logic is introduced in `build_and_run_sanitizer` function in `ps-common.sh`.
3. `jenkins-common.sh` contatins `filter_valgrind_log` and `setup_git_repo` which may be used by PXC scripts.


inikep:
I added comments to .sh files.

PR #463, opened 2025-05-26
PKG-473 [8.0]: Implement ccache to reduce compilation time by 50-80%

nogueiraanderson:
## PKG-473: Implement ccache for Jenkins builds to reduce compilation time

This PR adds ccache support to the Percona Server build pipeline, achieving **50-80% reduction in compilation time** by caching compilation artifacts in S3.

### Shared Library Integration
- Integrated with jenkins-pipelines shared library ([PR #3356](https://github.com/Percona-Lab/jenkins-pipelines/pull/3356))
- Uses `ccacheDownload` and `ccacheUpload` functions for S3 cache management
- Clean separation of concerns between build scripts and cache logic

### Smart Cache Key Generation
- **MySQL Version**: Extracted from sources (e.g., 8.0.42, 8.4.5)
- **Toolset**: Compiler + version (e.g., gcc-13.3.0, clang-18.1.8)
- **Build Configuration**: Including special builds (Debug, RelWithDebInfo, ASAN, Valgrind)
- **Build Parameters Type**: Distinguishes between standard/asan/valgrind/ubsan/msan builds

### Configuration Options
- **COMPILER**: Select specific compiler or use system default
- **CACHE_RETENTION_DAYS**: Configure S3 object lifecycle (7/14/21/30 days)
- **CCACHE_MAXSIZE**: Override default 8GB cache size limit
- Environment variables for compression, CPP2 mode, etc.

### Implementation Details
- **Jenkins Pipeline**: 
  - Added COMPILER parameter for toolset selection
  - Extract compiler version and create TOOLSET variable
  - Determine BUILD_PARAMS_TYPE based on ANALYZER_OPTS
- **Docker Build**:
  - Configure ccache environment variables
  - Mount .ccache directory for local cache
  - Use compression level 1 for optimal performance
- **Build Scripts**:
  - Only set CMake launcher options (no local ccache config)
  - Show cache statistics before/after builds

## Performance Results
- **First build**: Populates cache (no time savings)
- **Subsequent builds**: 50-80% faster compilation
- **Cache hit rate**: ~50% in testing (builds ~10,000 files)

## Credits
Implementation incorporates feedback from:
- @inikep (compiler detection, Azure patterns)
- @EvgeniyPatlan (shared library suggestion, shell fixes)  
- @vorsel (shared library review)
- @percona-ysorokin (cache key components, size recommendations)

## Related
- Jira: [PKG-473](https://perconadev.atlassian.net/browse/PKG-473)
- Shared Library PR: [jenkins-pipelines #3356](https://github.com/Percona-Lab/jenkins-pipelines/pull/3356)


jenkins/pipeline-parallel-mtr.groovy
```diff
@@ -199,6 +199,80 @@ void checkoutSources() {
     """
 }
 
+def downloadCcache() {
```
EvgeniyPatlan: maybe it is better to add this function as a shared library so it can be reused for other products?

jenkins/pipeline-parallel-mtr.groovy
```diff
@@ -199,6 +199,80 @@ void checkoutSources() {
...
+                echo "No existing cache, will create new one"
+            fi
+
+            # Set permissions
+            chmod -R 777 ${WORKSPACE}/.ccache || true
+        """
+    }
+}
+
+def uploadCcache() {
```
EvgeniyPatlan: maybe it is better to add this function as a shared library so it can be reused for other products?

jenkins/pipeline-parallel-mtr.groovy
vorsel: I was about to ask the same question)

nogueiraanderson: Noted. I'll implement it as a shared library.

jenkins/pipeline-parallel-mtr.groovy
nogueiraanderson: Done, @EvgeniyPatlan . See associated PR: https://github.com/Percona-Lab/jenkins-pipelines/pull/3356

local/build-binary
```diff
@@ -196,6 +227,12 @@ fi
 tar -C ${INSTALL_DIR}/usr/local --owner=0 --group=0 -czf "${WORKDIR}/$PRODUCT_FULL.tar.gz" $PRODUCT_FULL
 rm -rf ${INSTALL_DIR}
 
+# Show final ccache statistics
+if [[ "${USE_CCACHE}" == "1" ]] && command -v ccache &> /dev/null; then
```
EvgeniyPatlan:
```suggestion
if [[ "${USE_CCACHE}" == "1" ]] && [[ command -v ccache &> /dev/null ]]; then
```

local/build-binary
```diff
@@ -36,6 +36,37 @@ mkdir -p ${BUILD_DIR} ${INSTALL_DIR} ${DOWNLOAD_DIR}
...
 
 
+# ------------------------------------------------------------------------------
+# Setup ccache if enabled
+# ------------------------------------------------------------------------------
+if [[ "${USE_CCACHE}" == "1" ]]; then
+    echo "Setting up ccache..."
+
+    # Configure ccache (already installed in Docker image)
+    if command -v ccache &> /dev/null; then
```
EvgeniyPatlan, line 58:
```suggestion
    if [[ command -v ccache &> /dev/null ]]; then
```

jenkins/pipeline-parallel-mtr.groovy
```diff
@@ -611,7 +617,34 @@ pipeline {
...
+                        
+                        // Download ccache using shared library
+                        ccacheDownload([
+                            awsCredentialsId: AWS_CREDENTIALS_ID,
+                            s3Bucket: S3_ROOT_DIR + '/',
+                            workspace: env.WORKSPACE,
+                            branch: env.BRANCH,
+                            dockerOs: env.DOCKER_OS,
+                            cmakeBuildType: env.CMAKE_BUILD_TYPE,
+                            compiler: env.COMPILER,
```
inikep: The `COMPILER` is undefined by default in Jenkins.
To get best cache hits we need a full compiler version (e.g. `clang-18.1.8` or `gcc-13.3.0`)
In Azure Pipelines we use: https://github.com/percona/percona-server/blob/8.0/azure-pipelines.yml#L363-L367

jenkins/pipeline-parallel-mtr.groovy
```diff
@@ -611,7 +617,34 @@ pipeline {
                         git branch: JENKINS_SCRIPTS_BRANCH, url: JENKINS_SCRIPTS_REPO
 
                         checkoutSources()
+                        
+                        // Download ccache using shared library
+                        ccacheDownload([
+                            awsCredentialsId: AWS_CREDENTIALS_ID,
+                            s3Bucket: S3_ROOT_DIR + '/',
+                            workspace: env.WORKSPACE,
+                            branch: env.BRANCH,
```
inikep: The `PARENT_BRANCH` in Azure is misleading because it basically should be a server version (`5.7`, `8.0`, `8.4`). The idea behind it is that we want to share cache for the same server version even when a branch name is different.
Even better solution would be to distinguish `8.4.4`, `8.4.5` and so on:
We have this information in `MYSQL_VERSION` variable in https://github.com/Percona-Lab/ps-build/blob/8.0/local/build-binary#L126

jenkins/pipeline-parallel-mtr.groovy
inikep: Maybe with Jenkins it's not a big deal because we will change a compiler e.g. when docker images will be updated. With Azure we test many compilers with the same docker image.

inikep:
What about cache retention? Shouldn't we remove caches when they are not used/changed in 1 month?

docker/run-build
```diff
@@ -33,6 +33,14 @@ docker run --rm \
     --mount type=bind,source=${SOURCES_DIR},destination=/tmp/sources \
     --mount type=bind,source=${SCRIPTS_DIR},destination=/tmp/scripts \
     --mount type=bind,source=${WORK_DIR},destination=/tmp/work \
+    --mount type=bind,source=${ROOT_DIR}/.ccache,destination=/tmp/ccache \
+    -e CCACHE_DIR=/tmp/ccache \
+    -e USE_CCACHE=1 \
+    -e CCACHE_COMPRESS=1 \
+    -e CCACHE_COMPRESSLEVEL=9 \
+    -e CCACHE_CPP2=1 \
+    -e CCACHE_MAXSIZE=2G \
```
percona-ysorokin: 1. Not sure is 2GB is enough for our builds.
2. May be it is better to change persistent ccache options via command line (`ccache --max-size=8G`, for instance)

jenkins/pipeline-parallel-mtr.groovy
percona-ysorokin: Completely agree with @inikep here. Object names under which we upload ccache tarballs on AWS S3 should include at least the following 3 components:
1. \<mysql_version> (for instance, `8.0.42` or `8.4.5`)
2. \<toolset> (for instance, `clang-18.1.8` or `gcc-13.3.0`)
3. \<build_configuration> (`Debug` or `RelWithDebInfo`). Ideally ASan (`Debug` + `-DWITH_ASAN=ON`) and Valgrind (`Debug` + `-DWITH_Valgrind=ON`) should also have unique \<build_configuration> labels (e.g. `ASasn` and `Valgrind`).

local/build-binary
```diff
@@ -36,6 +36,37 @@ mkdir -p ${BUILD_DIR} ${INSTALL_DIR} ${DOWNLOAD_DIR}
...
+        ccache --version
+        ccache --max-size="${CCACHE_MAXSIZE:-2G}"
+        ccache --set-config compression=true
+        ccache --set-config compression_level=9
```
percona-ysorokin: Probably, no need to set up environment variables `CCACHE_COMPRESS=1` and `CCACHE_COMPRESSLEVEL=9` if you set these values here.

local/build-binary
```diff
@@ -36,6 +36,37 @@ mkdir -p ${BUILD_DIR} ${INSTALL_DIR} ${DOWNLOAD_DIR}
...
+# ------------------------------------------------------------------------------
+# Setup ccache if enabled
+# ------------------------------------------------------------------------------
+if [[ "${USE_CCACHE}" == "1" ]]; then
+    echo "Setting up ccache..."
+
+    # Configure ccache (already installed in Docker image)
+    if command -v ccache &> /dev/null; then
+        ccache --version
+        ccache --max-size="${CCACHE_MAXSIZE:-2G}"
```
percona-ysorokin: The same question regarding the 2GB size here.
For instance, the total size of all `.o` files in my clang-19 ASan build dir is 3.7 GB.
For gcc-14 ASan - 8GB.
May be with compression it will be smaller, but to be safe I would put 8GB limit there

percona-ysorokin:
> What about cache retention? Shouldn't we remove caches when they are not used/changed in 1 month?

I think it is possible to set this on S3 bucket / object level https://docs.aws.amazon.com/AmazonS3/latest/userguide/how-to-set-lifecycle-configuration-intro.html

inikep:
@nogueiraanderson My recent PR at https://github.com/Percona-Lab/ps-build/pull/465/files#diff-cbe31d84b2f944ad01d8f4adc62f6dc7303d467790017282389f014e4453734cR10 introduces `SERVER_VERSION` variable which you may use for `mysqlVersion`. I used `SERVER_VERSION` because it's also a version of Percona Server for MySQL.

nogueiraanderson:
## All Review Comments Addressed

All feedback has been addressed across multiple commits:

**@EvgeniyPatlan:**
- Shared library implementation: Created in jenkins-pipelines PR #3356
- Shell syntax: Fixed in commit c406c4c - now using double brackets with command substitution

**@inikep:**
- Compiler auto-detection: Added automatic detection when COMPILER is undefined (commit efb9767)
- Full compiler version: Using Azure Pipelines tested approach (commit ceb7732)
- SERVER_VERSION: Updated to use SERVER_VERSION instead of MYSQL_VERSION (commit db59262)
- Cache retention: Implemented via S3 lifecycle rules (PKG-769 completed)
- COMPILER parameter: Restored to Jenkins job definition (commit fc1a0ee)

**@percona-ysorokin:**
- Cache size: Increased to 8GB as suggested
- Build type suffixes: Added -asan/-valgrind suffixes for unique cache keys
- Environment variables: Removed redundant CCACHE_COMPRESS variables
- S3 lifecycle: Implemented tag-based retention (7, 14, 21, 30 days)

All commits are now visible in the PR.

**Related Jira:**
- [PKG-473](https://perconadev.atlassian.net/browse/PKG-473)
- [PKG-769](https://perconadev.atlassian.net/browse/PKG-769)

jenkins/pipeline-parallel-mtr.groovy
nogueiraanderson: Done. Thanks for the feedback @inikep and @percona-ysorokin!

local/build-binary
```diff
@@ -36,6 +36,37 @@ mkdir -p ${BUILD_DIR} ${INSTALL_DIR} ${DOWNLOAD_DIR}
...
+# ------------------------------------------------------------------------------
+# Setup ccache if enabled
+# ------------------------------------------------------------------------------
+if [[ "${USE_CCACHE}" == "1" ]]; then
+    echo "Setting up ccache..."
+
+    # Configure ccache (already installed in Docker image)
+    if [[ -n $(command -v ccache) ]]; then
+        ccache --version
+        ccache --max-size=8G
```
inikep: We can't change `ccache` settings in `local/build-binary` because when somebody will run this script locally without Docker it will change his/her `ccache`. I have `max_size = 256.0 GB` for example.

local/build-binary
nogueiraanderson: Thanks for the input, @percona-ysorokin .

jenkins/param-parallel-mtr.yml
```diff
@@ -47,11 +47,6 @@
         choices:
         - /usr/bin/cmake
         description: path to cmake binary
-    - choice:
-        name: COMPILER
-        choices:
-        - default
-        description: compiler version e.g. gcc-15 or clang-20
```
inikep: We shouldn't remove it (and keep `COMPILER`).

jenkins/pipeline-parallel-mtr.groovy
```diff
@@ -606,8 +621,97 @@ pipeline {
...
+                                elif [ -f sources/MYSQL_VERSION ]; then
+                                    cat sources/MYSQL_VERSION
+                                fi
+                            """).trim()
+
+                            if (versionFile) {
+                                def mysqlVersion = sh(returnStdout: true, script: """
+                                    echo "${versionFile}" | grep -E '^MYSQL_VERSION_MAJOR|^MYSQL_VERSION_MINOR|^MYSQL_VERSION_PATCH' | cut -d= -f2 | paste -sd'.' -
+                                """).trim()
+                                env.MYSQL_VERSION = mysqlVersion
```
inikep: All of it is already done with:
```
SERVER_VERSION = validatePsBranch()
```
at https://github.com/Percona-Lab/ps-build/blob/8.0/jenkins/pipeline-parallel-mtr.groovy#L585C28-L585C56
I'm using `SERVER_VERSION` because it's either version of MySQL or Percona Server for MySQL.

jenkins/pipeline-parallel-mtr.yml
```diff
@@ -43,6 +43,11 @@
         name: DOCKER_OS
         choices: '{docker_os_list}'
         description: OS version for compilation
+    - choice:
+        name: COMPILER
+        choices:
+        - default
+        description: compiler version (default = use system compiler)
```
inikep: COMPILER is already there (look several lines below).

docker/run-build
```diff
@@ -52,6 +53,15 @@ docker run --rm \
...
     export DIST_NAME='${DIST_NAME}'
     export SSL_VER='${SSL_VER}'
+
+    # ccache environment (ensure they're exported inside container)
+    export USE_CCACHE='${USE_CCACHE:-yes}'
+    export CCACHE_DIR='${CCACHE_DIR:-/tmp/ccache}'
+    export CCACHE_COMPRESS='${CCACHE_COMPRESS:-1}'
+    export CCACHE_COMPRESSLEVEL='${CCACHE_COMPRESSLEVEL:-1}'
+    export CCACHE_CPP2='${CCACHE_CPP2:-1}'
+    export CCACHE_MAXSIZE='8G'
```
inikep: This should depend on `CCACHE_MAXSIZE` from groovy file.

jenkins/param-parallel-mtr.yml
```diff
@@ -296,19 +294,3 @@
       - debian:bookworm
     jobs:
       - percona-server-{ps_version}-param-parallel-mtr
-
-- project:
-    name: param-parallel-mtr-9.x
```
inikep: This reverts `jenkins/param-parallel-mtr.yml` to some state in the past or 8.x branch. 
Probably you shouldn't touch this file at all.

inikep (approved):
LGTM

percona-ysorokin (approved):
LGTM

PR #466, opened 2025-06-03
PS-9821 [8.0]: Improve docker images

inikep:
1. Switch default `gcc` from `gcc-9` to `gcc-10` for `ubuntu-focal` Docker image. This fixes an issue with reading a bad compiler version for `ccache`. It also fixes the issue that `COMPILER` was always ignored because it was overwritten by `CMAKE_OPTS+=" -DCMAKE_C_COMPILER=gcc-10 -DCMAKE_CXX_COMPILER=g++-10"`.
2. Change boost downloading repo to `archives.boost.io` because of issues with `downloads.sourceforge.net` from Europe (Hetzner cloud).


vorsel (approved):

percona-ysorokin (approved):
LGTM

PR #469, opened 2025-06-25
PKG-473: Fix Jenkins pipeline Groovy syntax error

nogueiraanderson:
### Problem
The merged PR #463 introduced a syntax error in the Jenkins pipeline that caused build failures:
```
WorkflowScript: 707: Expected a step @ line 707, column 25.
def retentionDays = (env.BUILD_PARAMS_TYPE in ['asan', 'valgrind']) ?
```

### Root Cause
The `in` operator cannot be used directly in Jenkins pipeline scripts. This is a Groovy language limitation within the Jenkins pipeline context.

### Solution
Replace the `in` operator with standard comparison operators:
```groovy
// Before (causes error):
def retentionDays = (env.BUILD_PARAMS_TYPE in ['asan', 'valgrind']) ? 

// After (fixed):
def retentionDays = (env.BUILD_PARAMS_TYPE == 'asan' || env.BUILD_PARAMS_TYPE == 'valgrind') ?
```

### Testing
- Ran `npm-groovy-lint` to verify no syntax errors remain

### Related
- Original PR: #463

PR #471, opened 2025-06-26
PKG-473: Additional pipeline fixes for ccache implementation

nogueiraanderson:
## Summary
This PR contains the remaining fix needed for the ccache implementation to work correctly in Jenkins.

## Issue Fixed
**Missing property error**: "No such property: CCACHE_MAXSIZE" - Fixed by setting it as environment variable

## Changes
- Set  as an environment variable using the idiomatic  pattern
- This allows the constant defined at the top of the pipeline to be accessible in shell scripts

## Testing
- Tested on https://ps80.cd.percona.com/job/PKG-473-ccache-test/
- Build completed successfully with ccache working (5,120 files cached, 0.51GB cache size)
- Error is resolved

## Related
- Original implementation: #463 (merged June 25)
- First fix: #469 (merged June 26)
- This PR contains the additional fix discovered during testing
- Jira ticket: PKG-473

### DIFF
--- a/jenkins/pipeline-parallel-mtr.groovy
+++ b/jenkins/pipeline-parallel-mtr.groovy
@@ -240,7 +240,7 @@ void build(String SCRIPT) {
                     if [ \$(docker ps -q | wc -l) -ne 0 ]; then
                         docker ps -q | xargs docker stop --time 1 || :
                     fi
-                    eval USE_CCACHE=yes CCACHE_MAXSIZE=${CCACHE_MAXSIZE} KEEP_BUILD=yes ${SCRIPT} ${DOCKER_OS} ${WORKSPACE}/${WORK_DIR}
+                    eval USE_CCACHE=yes CCACHE_MAXSIZE=${env.CCACHE_MAXSIZE} KEEP_BUILD=yes ${SCRIPT} ${DOCKER_OS} ${WORKSPACE}/${WORK_DIR}
                 " 2>&1 | tee build.log
 
                 echo Archive build log: \$(date -u "+%s")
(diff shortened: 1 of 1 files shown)


inikep:
`WORK_DIR` used in the same command and also declared at top works fine: 
```
WORK_DIR = 'work'
```

From ChatGPT:
Groovy tries to resolve `${CCACHE_MAXSIZE}` before passing the string to the shell.
If it's declared `final`, and used inside a closure or nested scope, Groovy might not see it as a property (hence the "No such property" error).

nogueiraanderson:
@inikep  Thanks for looking into this! I tested your suggestion but removing `final` alone didn't resolve the issue - the build still failed with the same error.

The problem is deeper than just the final modifier. In Jenkins pipelines, variables declared at the script level aren't accessible inside shell script interpolations within functions, regardless of how they're declared. This is why `WORK_DIR` works in some places but `CCACHE_MAXSIZE` fails here.


```bash
[Pipeline] // script
[Pipeline] }
[Pipeline] // stage
[Pipeline] }
[Pipeline] // timeout
[Pipeline] }
[Pipeline] // node
[Pipeline] End of Pipeline
Also:   org.jenkinsci.plugins.workflow.actions.ErrorAction$ErrorId: 1b813eaf-4331-4e4c-acce-13233103a103
groovy.lang.MissingPropertyException: No such property: CCACHE_MAXSIZE for class: groovy.lang.Binding
	at groovy.lang.Binding.getVariable(Binding.java:63)
```

inikep (approved):
LGTM

EvgeniyPatlan (approved):

PR #472, opened 2025-07-03
PKG-473: Fix Docker mount dependency issue for ccache directory

nogueiraanderson:
## Summary
Fixes Docker mount failure when .ccache directory doesn't exist.

## Problem
Docker script assumes .ccache directory exists but dependency can be violated:
- Jenkins creates `${env.WORKSPACE}/.ccache` 
- Docker expects `${ROOT_DIR}/.ccache`
- Path mismatch or creation failure causes mount error

## Solution
```bash
mkdir -p "${ROOT_DIR}/.ccache"
```

Always create directory before Docker mount - simple, low-risk approach.

PR #473, opened 2025-07-03
PKG-473: Fix ccache permissions inside Docker container

nogueiraanderson:
## Summary
Fixes ccache permission issues inside Docker container by adding `/tmp/ccache` to chown commands.

## Problem
After PR #472 fixed the directory creation issue, another problem can occur:
- Docker mounts `.ccache` at `/tmp/ccache` 
- The mysql user needs write permissions during build
- Without proper permissions, ccache fails with "permission denied" errors

## Solution
Add `/tmp/ccache` to both chown commands in `docker/run-build`:
1. Before build: `chown mysql:mysql ... /tmp/ccache`
2. After build: `chown $(id -u):$(id -g) ... /tmp/ccache`

This ensures ccache can write cache files properly during the build process.

## Related to
- PR #472 (creates .ccache directory)
- PKG-473 (ccache implementation)

### DIFF
--- a/docker/run-build
+++ b/docker/run-build
@@ -68,11 +68,11 @@ docker run --rm \
     export DOCKER_OS='${SOURCE_IMAGE}'
     export KEEP_BUILD='${KEEP_BUILD}'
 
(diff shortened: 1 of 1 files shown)


inikep (approved):
LGTM

percona-ysorokin (approved):
LGTM

PR #477, opened 2025-08-20
PS-10061: Add support for PS 9.4.0 in suites-groups.sh

inikep:
Support a different list of MTR suites for PS 9.4.0 (`jdv` added) and older.


percona-ysorokin (approved):
LGTM

vorsel (approved):

PR #479, opened 2025-08-22
PS-10120: Improve `percona-server-x.y-param-parallel-mtr` and `percona-server-x.y-pipeline-parallel-mtr` pipelines

inikep:
1. Fix missing `#!/bin/bash` in `env.COMPILER_VERSION` (PKG-1061)
2. Check if `sources` exists and is a valid git repo with a proper permissions. In case repo is not valid it's removed as it probably comes from a previous run which was not cleaned properly.
3.  `ALLOW_ABORTED_WORKERS_RERUN` was created for spot instances that were killed by AWS. But this option created a side-effect that all pipeline FAILUREs are converted to UNSTABLEs. Fixes:
a) disable `ALLOW_ABORTED_WORKERS_RERUN` by default
b) use `catchError(buildResult: 'UNSTABLE')` only if `ALLOW_ABORTED_WORKERS_RERUN` is enabled
c) add `try {} catch (err)` to print a warning that a worker/agent failed

### DIFF
--- a/jenkins/pipeline-parallel-mtr.groovy
+++ b/jenkins/pipeline-parallel-mtr.groovy
@@ -661,7 +661,7 @@ pipeline {
                             // Extract compiler version for ccache key
                             def CC_COMPILER = env.CC ?: 'gcc'
 
-                            env.COMPILER_VERSION = sh(returnStdout: true, script: """
+                            env.COMPILER_VERSION = sh(returnStdout: true, script: """#!/bin/bash
                                 COMPILER="${CC_COMPILER}"
                                 if [[ "\$COMPILER" == *"clang"* ]]; then
                                     \$COMPILER --version | grep -o "clang version.*" | awk '{print \$3}'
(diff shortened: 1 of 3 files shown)


nogueiraanderson (approved):

nogueiraanderson (approved):
LGTM!

inikep:
Tested at https://ps80.cd.percona.com/job/percona-server-TEST-pipeline-parallel-mtr/1/

jenkins/pipeline-parallel-mtr.groovy
```diff
@@ -100,10 +100,22 @@ void prepareWorkspace(Integer WORKER_ID) {
...
+                if [ -d sources ]; then
+                    sudo cat sources/MYSQL_VERSION || :
+                    if ! sudo git -C sources status >/dev/null 2>&1; then
+                        echo "Warning: Git repo in ./sources is broken, removing it. It should happen only for re-runs."
+                        sudo GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null git -C sources -c safe.directory="$WORKSPACE/sources" log --oneline -10 || :
+                        sudo rm -rf sources
+                    else
+                        sudo git -C sources log --oneline -10
+                        sudo git -C sources reset --hard
+                        sudo git -C sources clean -xdf
```
percona-ysorokin, line 114: May be also `git -C sources submodule update --init` just in case

percona-ysorokin (approved):

jenkins/pipeline-parallel-mtr.groovy
inikep: Done

PR #483, opened 2025-09-16
PS-10105: Support building and running test suite for JS Stored Routi…

dlenev:
…nes in Jenkins (part 1)

https://perconadev.atlassian.net/browse/PS-10105

Add support for building Percona Server with JS Stored Programs component.

Such build can be enabled by turning ON newly introduced WITH_JS_LANG Jenkins parameter (which is OFF by default). To support this build scripts were modified to automatically download and use appropriate version of V8 engine static library and headers.

Note that we don't enable running component_js_lang test suite by default yet.

### DIFF
--- a/docker/run-build
+++ b/docker/run-build
@@ -50,6 +50,7 @@ docker run --rm \
     export WITH_ROUTER='${WITH_ROUTER}'
     export WITH_GCOV='${WITH_GCOV}'
     export WITH_MYSQLX='${WITH_MYSQLX}'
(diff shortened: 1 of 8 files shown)


dlenev:
Here are results of testing the above changes on Jenkins (new test pipeline which copied pipeline-parallel-mtr was created to test them) - https://ps80.cd.percona.com/job/percona-server-8.x-pipeline-parallel-mtr-js-lang/

inikep (approved):
LGTM, good job

kamil-holubicki (approved):

PR #487, opened 2025-10-10
PS-10223: Add ARM platforms to Jenkins

dlenev:
https://perconadev.atlassian.net/browse/PS-10223

Extended prepare-ps-build-docker pipeline and supporting script so it can now prepare ps-build docker images, which are used for PS build and test runs in Jenkins, for ARM64 platform.

These images are uploaded to ECR with -aarch64 suffix in their tag so they can be correctly identified when used from pipeline-parallel-mtr-*.* pipelines.

Added new ARCH parameter to pipeline-parallel-mtr-*.* pipelines, which allows to choose whether we want to build and test on x86_64 (default) or aarch64 (ARM64) architecture. Adjusted scripts for building server and running tests to take this parameter/the fact that they can be run on ARM64 into account.

Added new ARCH axis to matrix parameter of param-parallel-mtr-*.* job. Combinations using aarch64 (ARM64) architecture are not enabled by default.

### DIFF
--- a/jenkins/param-parallel-mtr.yml
+++ b/jenkins/param-parallel-mtr.yml
@@ -25,7 +25,7 @@
     - matrix-combinations:
         name: COMBINATIONS
         description: "Select matrix combinations"
-        filter: CMAKE_BUILD_TYPE=="Debug" && DOCKER_OS=="ubuntu:noble" || CMAKE_BUILD_TYPE=="RelWithDebInfo" && DOCKER_OS=="ubuntu:noble" || CMAKE_BUILD_TYPE=="Debug" && DOCKER_OS=="debian:bookworm" || CMAKE_BUILD_TYPE=="RelWithDebInfo" && DOCKER_OS=="debian:bookworm" || CMAKE_BUILD_TYPE=="Debug" && DOCKER_OS=="oraclelinux:9" || CMAKE_BUILD_TYPE=="RelWithDebInfo" && DOCKER_OS=="oraclelinux:9"
+        filter: CMAKE_BUILD_TYPE=="Debug" && DOCKER_OS=="ubuntu:noble" && ARCH=="x86_64" || CMAKE_BUILD_TYPE=="RelWithDebInfo" && DOCKER_OS=="ubuntu:noble" && ARCH=="x86_64" || CMAKE_BUILD_TYPE=="Debug" && DOCKER_OS=="debian:bookworm" && ARCH=="x86_64" || CMAKE_BUILD_TYPE=="RelWithDebInfo" && DOCKER_OS=="debian:bookworm" && ARCH=="x86_64" || CMAKE_BUILD_TYPE=="Debug" && DOCKER_OS=="oraclelinux:9" && ARCH=="x86_64" || CMAKE_BUILD_TYPE=="RelWithDebInfo" && DOCKER_OS=="oraclelinux:9" && ARCH=="x86_64"
     - string:
         name: GIT_REPO
         default: "https://github.com/percona/percona-server"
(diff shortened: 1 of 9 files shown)


dlenev:
Test of changes in prepare-ps-build-docke pipeliner: https://ps80.cd.percona.com/job/prepare-ps-build-docker-arm64/5/
Test of changes in pipeline-parallel-mtr pipelines: https://ps80.cd.percona.com/job/percona-server-8.0-pipeline-parallel-mtr-arm64/
Test of changes in param-parallel-mtr job: https://ps80.cd.percona.com/job/percona-server-8.0-param-parallel-mtr-arm64/

jenkins/prepare-ps-build-docker.groovy
```diff
@@ -0,0 +1,56 @@
+library changelog: false, identifier: 'lib@master', retriever: modernSCM([
+    $class: 'GitSCMSource',
+    remote: 'https://github.com/Percona-Lab/jenkins-pipelines.git'
+])
```
inikep: `jenkins-pipelines.git` was required only for `ccache` functions. Probably we don't need it here.

jenkins/param-parallel-mtr.yml
```diff
@@ -25,7 +25,7 @@
     - matrix-combinations:
         name: COMBINATIONS
         description: "Select matrix combinations"
-        filter: CMAKE_BUILD_TYPE=="Debug" && DOCKER_OS=="ubuntu:noble" || CMAKE_BUILD_TYPE=="RelWithDebInfo" && DOCKER_OS=="ubuntu:noble" || CMAKE_BUILD_TYPE=="Debug" && DOCKER_OS=="debian:bookworm" || CMAKE_BUILD_TYPE=="RelWithDebInfo" && DOCKER_OS=="debian:bookworm" || CMAKE_BUILD_TYPE=="Debug" && DOCKER_OS=="oraclelinux:9" || CMAKE_BUILD_TYPE=="RelWithDebInfo" && DOCKER_OS=="oraclelinux:9"
+        filter: CMAKE_BUILD_TYPE=="Debug" && DOCKER_OS=="ubuntu:noble" && ARCH=="x86_64" || CMAKE_BUILD_TYPE=="RelWithDebInfo" && DOCKER_OS=="ubuntu:noble" && ARCH=="x86_64" || CMAKE_BUILD_TYPE=="Debug" && DOCKER_OS=="debian:bookworm" && ARCH=="x86_64" || CMAKE_BUILD_TYPE=="RelWithDebInfo" && DOCKER_OS=="debian:bookworm" && ARCH=="x86_64" || CMAKE_BUILD_TYPE=="Debug" && DOCKER_OS=="oraclelinux:9" && ARCH=="x86_64" || CMAKE_BUILD_TYPE=="RelWithDebInfo" && DOCKER_OS=="oraclelinux:9" && ARCH=="x86_64"
```
inikep: I would suggest to create separate pipelines for `arm64` (for example `percona-server-8.0-arm64-param`/`parallel`, we don't need `parallel-mtr` anymore) . The reasons are:
1. The new matrix looks complicated.
2. The jobs / results for x86_64/aarch64 will be mixed. Currently our ASan/Valgrind jobs are separated and it's easy to compare results even after 3 months. `aarch64` will be hidden between `x86_64` results which I expect to be run more frequent.

jenkins/prepare-ps-build-docker.groovy
dlenev: Indeed. Fixed.

jenkins/param-parallel-mtr.yml
dlenev: As discussed on Zoom I would still prefer to have separate pipelines for aarch64 but still invoke them from the same matrix jobs as x86_64 pipelines. However, in practice, it turned to be rather complicated for me to achieve.

So I have created both separate matrix and pipeline jobs for aarch64 as you have suggested initially.

dlenev: After another round of discussion with @inikep we decided to keep the initial approach with single matrix and single pipeline job per version.

inikep (approved):
LGTM

vorsel (approved):

percona-ysorokin (approved):
LGTM

PR #496, opened 2025-10-30
PKG-1161: Fix "Could not get lock /var/lib/apt/lists/lock" on Hetzner instances

inikep:
1. Fix "Could not get lock /var/lib/apt/lists/lock" on Hetzner instances:
- Introduce `apt_get_retry()` and `yum_retry()` functions to retry every second
2. Print system info for each worker

### DIFF
--- a/local/utils.inc.sh
+++ b/local/utils.inc.sh
@@ -148,6 +148,25 @@ function setup_git_repo() {
     popd
 }
 
+# usage: apt_get_retry <list of commands>
+function apt_get_retry() {
+  if [ -f /usr/bin/apt ]; then
+    until sudo DEBIAN_FRONTEND=noninteractive apt-get $@; do
+      echo "apt-get failed, retrying in 1 second..."
+      sleep 1
+    done
+  fi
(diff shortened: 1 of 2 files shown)


inikep:
Tested at https://ps80.cd.percona.com/view/ASan+Valgrind/job/percona-server-TEST-pipeline-parallel-mtr/77

jakub-nowakowski-percona (approved):

local/utils.inc.sh
```diff
@@ -148,6 +148,25 @@ function setup_git_repo() {
     popd
 }
 
+# usage: apt_get_retry <list of commands>
+function apt_get_retry() {
```
kamil-holubicki, line 152: I recommend adding some timeout for this loop to avoid infinite looping.

kamil-holubicki (approved):

local/utils.inc.sh
inikep: When I will find a job that stops at the loop I will add this feature.

PR #499, opened 2025-11-20
PS-10183: Install `gcc-toolset-14` for `centos:8` image

dlenev:
Upstream has switched to `gcc-toolset-14` for RHEL-8 compatible distros in mysql-9.5.0.


dlenev:
Here is results of test run for the above changes: https://ps80.cd.percona.com/job/prepare-ps-build-docker-dlenev/1/

EvgeniyPatlan (approved):

inikep (approved):
LGTM

PR #501, opened 2025-11-22
PS-10183: Set USER env variable before running MTR to fix router.dest…

dlenev:
…ination_socket failures in Jenkins.

This test expects USER environment variable to contain name of system user under which tests are executed. And normally, it is the case, as it is set accordingly by 'login' process. However, this variable is not set in Docker environment by default.

Fix test failures in Jenkins caused by this by explicitly setting USER environment variable before running MTR.


percona-ysorokin (approved):
LGTM

inikep (approved):
LGTM

PR #503, opened 2026-01-22
PS-10398 [8.0]: Always wipe WORK_DIR to avoid stale work/extract on reruns

inikep:
1. `git clean -xdf` sometimes leaves work/extract behind (e.g. “Directory not empty”), which can cause later MTR workers to see stale/partial extraction state and fail with missing files (like `mysql-test-run.pl`).
Add an explicit `sudo rm -rf ${WORK_DIR}` cleanup in `prepareWorkspace()` so reruns start from a clean `work/` directory.
2. Add `set -euo pipefail` to detect failures in `./docker/run-build`.

### DIFF
--- a/jenkins/pipeline-parallel-mtr.groovy
+++ b/jenkins/pipeline-parallel-mtr.groovy
@@ -100,6 +100,16 @@ void prepareWorkspace(Integer WORKER_ID, boolean UNIT_TESTS) {
             sudo git log --oneline -10
             sudo git reset --hard
 
+            wipe_work_dir() {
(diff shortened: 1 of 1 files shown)


inikep:
Tested at https://ps80.cd.percona.com/job/percona-server-TEST-pipeline-parallel-mtr/86/console (intentional build error)
and https://ps80.cd.percona.com/job/percona-server-TEST-pipeline-parallel-mtr/85/console

vorsel (approved):

percona-ysorokin (approved):
LGTM

PR #504, opened 2026-02-11
Fail in case version file can't be donwloaded. This is an indication …

percona-mhansson:
…that

something went seriously wrong and any subsequent errors will point in the wrong direction.

### DIFF
--- a/jenkins/pipeline-parallel-mtr.groovy
+++ b/jenkins/pipeline-parallel-mtr.groovy
@@ -370,11 +370,18 @@ def getServerVersion() {
                 GIT_REPO_LINK=$(echo ${GIT_REPO} | sed -e "s|github|x-access-token:${JNKPercona_token}@github|g")
             fi
             RAW_VERSION_LINK=$(echo ${GIT_REPO_LINK%.git} | sed -e "s:github.com:raw.githubusercontent.com:g")
-            REPLY=$(curl -Is ${RAW_VERSION_LINK}/${BRANCH}/MYSQL_VERSION | head -n 1 | awk '{print $2}')
-            if [[ ${REPLY} != 200 ]]; then
-                curl ${RAW_VERSION_LINK}/${BRANCH}/VERSION -o ${WORKSPACE}/VERSION-${BUILD_NUMBER}
-            else
-                curl ${RAW_VERSION_LINK}/${BRANCH}/MYSQL_VERSION -o ${WORKSPACE}/VERSION-${BUILD_NUMBER}
+
+            for VERSION_FILE in MYSQL_VERSION VERSION; do
+                HTTP_CODE=$(curl ${RAW_VERSION_LINK}/${BRANCH}/${VERSION_FILE} -w "%{http_code}" \
+                    -o ${WORKSPACE}/VERSION-${BUILD_NUMBER}
+                if [ "$HTTP_CODE" = "200" ]; then
+                    break
+                fi
+            done
+
+            if [[ ! -s ${WORKSPACE}/VERSION-${BUILD_NUMBER} ]]; then
+                echo "Failed to download ${VERSION_FILE} or it's empty"
+                exit 1
             fi
          '''
     } // withCredentials
@@ -749,7 +756,7 @@ pipeline {
                         script {
                             // Set ccache size as environment variable
                             env.CCACHE_MAXSIZE = CCACHE_MAXSIZE
-                            
+
                             // Set BUILD_PARAMS_TYPE based on ANALYZER_OPTS
                             if (env.ANALYZER_OPTS) {
                                 if (env.ANALYZER_OPTS.contains('ASAN')) {


jenkins/pipeline-parallel-mtr.groovy
```diff
@@ -392,10 +372,12 @@ def getServerVersion() {
...
             REPLY=$(curl -Is ${RAW_VERSION_LINK}/${BRANCH}/MYSQL_VERSION | head -n 1 | awk '{print $2}')
             if [[ ${REPLY} != 200 ]]; then
-                curl ${RAW_VERSION_LINK}/${BRANCH}/VERSION -o ${WORKSPACE}/VERSION-${BUILD_NUMBER}
+                VERSION_FILE=VERSION
             else
-                curl ${RAW_VERSION_LINK}/${BRANCH}/MYSQL_VERSION -o ${WORKSPACE}/VERSION-${BUILD_NUMBER}
+                VERSION_FILE=MYSQL_VERSION
             fi
+            curl --fail ${RAW_VERSION_LINK}/${BRANCH}/${VERSION_FILE} -o ${WORKSPACE}/VERSION-${BUILD_NUMBER} || \
+                { echo "Failed to download MYSQL_VERSION file"; exit 1; }
```
inikep: Failed to download ${VERSION_FILE}  file"

percona-mhansson: Done.

I also added some more tests. In the case when the URL doesn't contain `https://`, the file is empty if not found (regardless of `--fail`). If you use `https://` but omit the `--fail` flag, the HTTP error ends up in the file. Just in case that the latter can happen under other circumstances, I added tests for both.

jenkins/pipeline-parallel-mtr.groovy
```diff
@@ -392,9 +372,19 @@ def getServerVersion() {
...
-                curl ${RAW_VERSION_LINK}/${BRANCH}/MYSQL_VERSION -o ${WORKSPACE}/VERSION-${BUILD_NUMBER}
+                VERSION_FILE=MYSQL_VERSION
+            fi
+            curl --fail ${RAW_VERSION_LINK}/${BRANCH}/${VERSION_FILE} -o ${WORKSPACE}/VERSION-${BUILD_NUMBER} || \
+                { echo "Failed to download ${VERSION_FILE} file"; exit 1; }
+            if [[ ! -s ${WORKSPACE}/VERSION-${BUILD_NUMBER} ]]; then
+                echo "Failed to download ${VERSION_FILE} or it's empty"
+                exit 1
+            fi
+            if grep -q "^404" ${VERSION_FILE}; then
```
inikep: This will not work. Local file is `${WORKSPACE}/VERSION-${BUILD_NUMBER}`

FROM CHATGPT:
Cleaner and safer version:
- Tries MYSQL_VERSION first
- Falls back to VERSION
- Uses HTTP status directly from curl
- Validates the downloaded file correctly

```
VERSION_FILE=MYSQL_VERSION
TARGET="${WORKSPACE}/VERSION-${BUILD_NUMBER}"

HTTP_CODE=$(curl -s -o /dev/null -w "%{http_code}" \
    "${RAW_VERSION_LINK}/${BRANCH}/${VERSION_FILE}")

if [[ "${HTTP_CODE}" != "200" ]]; then
    VERSION_FILE=VERSION
fi

curl -sSf "${RAW_VERSION_LINK}/${BRANCH}/${VERSION_FILE}" -o "${TARGET}" || {
    echo "Failed to download ${VERSION_FILE} file"
    exit 1
}

if [[ ! -s "${TARGET}" ]]; then
    echo "Downloaded ${VERSION_FILE} is empty"
    exit 1
fi

# Optional: guard against HTML error pages
if grep -qiE '<html|not found|404' "${TARGET}"; then
    echo "Downloaded file appears to be an error page"
    exit 1
fi

```

percona-mhansson: I wrote it with Copilot 😂 

A little bit skeptical about `-o /dev/null`, though, as this (potentially) downloads the file, throws it away, and then downloads it again.

percona-mhansson: I think we have the ultimate version now! 💪

inikep:
If you will merge this PR it will change pipelines for 5.7 (parallel), 8.0, 8.4, and 9.7.
Without testing we had to revert some changes in the past.
It would be good if you would test your changes on a temp pipeline.

inikep (approved):
LGTM

inikep:
I created a pipeline at https://ps80.cd.percona.com/job/percona-server-TEST-mhansson-pipeline-parallel-mtr/
that uses https://github.com/percona-mhansson/ps-build for testing

inikep:
@percona-mhansson My new PR improves error reporting and also improves on top of your changes:
https://github.com/Percona-Lab/ps-build/pull/505/changes/46c6d0d17a1479c5f8812df8e4772c81d26a502e

PR #506, opened 2026-02-17
PS-10408: Fix tar failure after EXTRACT_DIR removal by restoring cwd

inikep:
When `KEEP_BUILD!=yes`, `rm -rf ${EXTRACT_DIR}` deletes the shell's current working directory. The subsequent tar command then fails with `Cannot getcwd: No such file or directory`. Add `cd ${WORKDIR}` before tar to move into a directory that is guaranteed to still exist.

### DIFF
--- a/local/test-binary-parallel-mtr
+++ b/local/test-binary-parallel-mtr
@@ -640,6 +640,7 @@ if [[ "${KEEP_BUILD}" != "yes" ]]; then
     rm -rf ${EXTRACT_DIR}
 fi
 
+cd ${WORKDIR}
 tar -C ${RESULTS_DIR} --owner=0 --group=0 -czf "${RESULTS_DIR}/ps80-test-mtr_logs-${WORKER_NO}.tar.gz" mtr_var
 
 ls -l ${WORKDIR}


nogueiraanderson (approved):

local/test-binary-parallel-mtr
```diff
@@ -640,6 +640,7 @@ if [[ "${KEEP_BUILD}" != "yes" ]]; then
     rm -rf ${EXTRACT_DIR}
 fi
 
+cd ${WORKDIR}
```
percona-mhansson, line 643: I don't really see the problem? Your tar command starts with `-C` which changes the directory for the duration of that command, so there shouldn't be any problem afaict. Are you saying tar fails because the shell's current directory doesn't exist? Are some of the paths relative?

inikep: The issue: the script's cwd was inside ${EXTRACT_DIR} (i.e. /tmp/work/extract/mysql-test). When `rm -rf ${EXTRACT_DIR}` on line 640 deletes `/tmp/work/extract`, the shell's current directory ceases to exist. The subsequent tar command then fails with `Cannot getcwd: No such file or directory` because the OS can no longer resolve the process's working directory.

percona-mhansson: Are you sure that's the problem? I experimented with doing that on both bash and zsh and it works fine. (standing in a non-existing directory). But sure, if it solves the problem...

percona-mhansson (approved):

PR #512, opened 2026-03-12
PKG-1279: enable ccache on Hetzner with USE_CCACHE parameter

nogueiraanderson:
Fixes PKG-1279: ccache on Hetzner (Jenkins) doesn't work.

ccache on Hetzner was never properly implemented in the pipeline. Two issues:

1. **`USE_CCACHE` parameter missing from JJB template.** Never added to the deployed job config. At runtime `params.USE_CCACHE` is null, so ccache is always disabled ("ccache disabled" in build logs).

2. **Wrong credentials passed to shared library on Hetzner.** The pipeline always passes `awsCredentialsId: AWS_CREDENTIALS_ID` regardless of cloud. The shared library's Hetzner override was dead code (fixed in jenkins-pipelines#3876).

**Changes:**
- Normalize `env.USE_CCACHE = params.USE_CCACHE ?: 'yes'` so ccache works even before JJB deploys the param
- Pass `HTZ_STASH` and `s3://percona-jenkins-artifactory/` when building on Hetzner
- Add `USE_CCACHE` choice parameter to both 8.0 and 5.7 JJB templates
- Include `USE_CCACHE` in aborted-worker rerun parameters

Depends on Percona-Lab/jenkins-pipelines#3876 (merge first).

### DIFF
--- a/jenkins/pipeline-57-parallel-mtr.yml
+++ b/jenkins/pipeline-57-parallel-mtr.yml
@@ -94,6 +94,13 @@
         default:
         description: make options, like VERBOSE=1
     - choice:
+        name: USE_CCACHE
+        choices:
+        - "yes"
+        - "no"
+        default: "yes"
+        description: Use ccache to speed up compilation (80-90% faster rebuilds)
+    - choice:
         name: CI_FS_MTR
         choices:
         - "yes"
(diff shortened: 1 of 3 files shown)


kaushikpuneet07 (approved):
LGTM

PR #515, opened 2026-05-21
PS-10960 Update the pattern to match the new eol version format

VarunNagaraju:
https://perconadev.atlassian.net/browse/PS-10960

The new pattern will also match the old version properly.


inikep (approved):
LGTM

PR #518, opened 2026-05-28
PS-11179: Wire pipeline-parallel-mtr to resolveArmWorker (Hetzner to AWS Graviton fallback)

nogueiraanderson:
**Feature**
- Routes the 12 PS MTR jobs (9 on ps80, 3 on ps57) through `vars/resolveArmWorker` so `CLOUD=auto` falls back from Hetzner CAX (arm64) to AWS Graviton when capacity is unavailable. Explicit `Hetzner`/`AWS`/`epyc9654` callers see no behaviour change (resolver defaults match the existing label switch one-to-one across all 8 `CLOUD x ARCH` combinations).

**Why**
- Hetzner ARM (CAX) intermittently unavailable since [PKG-1305](https://perconadev.atlassian.net/browse/PKG-1305); only workaround today is a manual `CLOUD=AWS` re-run.
- Substrate already live: resolver in [PR 4125](https://github.com/Percona-Lab/jenkins-pipelines/pull/4125), Graviton Fleet ASGs applied on ps3 + ps57 via [cd PR 18](https://github.com/percona/percona-cd-platform/pull/18), `EC2FleetCloud` codified in [PR 4126](https://github.com/Percona-Lab/jenkins-pipelines/pull/4126). End-to-end smoke validated on both masters.
- Single shared groovy reaches all 12 leaves. The 6 downstream `params.CLOUD` reads in `ccacheDownload`/`ccacheUpload` swap to `CLOUD_CHOSEN` together: that argument drives the S3 endpoint URL (Hetzner FSN1 vs AWS) AND the `put-object-tagging` skip inside the shared lib, so partial application would yield AWS creds with a Hetzner endpoint.

**Tickets**
- [PS-11179](https://perconadev.atlassian.net/browse/PS-11179) (this PR)
- Staged rollout follow-ups (separate PRs once canary green): ps57 default flip, ps80 default flip.

### DIFF
--- a/jenkins/pipeline-57-parallel-mtr.yml
+++ b/jenkins/pipeline-57-parallel-mtr.yml
@@ -134,7 +134,8 @@
         - "Hetzner"
         - "AWS"
         - "epyc9654"
-        description: "Host provider for Jenkins workers"
+        - "auto"
+        description: "Host provider for Jenkins workers. `auto` routes arm64 builds to AWS Graviton when Hetzner ARM (CAX) capacity is unavailable; x86 always stays on Hetzner."
     - choice:
         name: FULL_MTR
         choices:
--- a/jenkins/pipeline-parallel-mtr.groovy
+++ b/jenkins/pipeline-parallel-mtr.groovy
@@ -692,12 +692,12 @@ def notifySlack(status, color, customMessage) {
 // than the old silent-AWS fallback; choice-param validation enforces the
 // list at every form/REST entry point so this is reachable only via auto.
 def resolved = resolveArmWorker(cloud: params.CLOUD, arch: params.ARCH)
-LABEL        = resolved.label
-MICRO_LABEL  = resolved.microLabel
-CLOUD_CHOSEN = resolved.cloudChosen
+LABEL        = resolved?.label
+MICRO_LABEL  = resolved?.microLabel
+CLOUD_CHOSEN = resolved?.cloudChosen
 resolved     = null  // drop map reference for CPS friendliness
 if (!LABEL || !MICRO_LABEL || !CLOUD_CHOSEN) {
-    error("resolveArmWorker returned invalid result: " +
+    error("resolveArmWorker returned invalid result (or null): " +
           "LABEL=${LABEL} MICRO_LABEL=${MICRO_LABEL} CLOUD_CHOSEN=${CLOUD_CHOSEN}")
 }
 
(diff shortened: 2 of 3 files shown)


nogueiraanderson:
#### Canary protocol (between PR A merge and ps57 default flip)

1. **Healthy-Hetzner path**: trigger one ps57 MTR job (e.g. `percona-server-5.7-pipeline-parallel-mtr`) with `CLOUD=auto`, `ARCH=aarch64`. Confirm a Hetzner CAX worker provisions; `currentBuild.description` reads `[Hetzner:docker-aarch64]`.
2. **Fallback path**: write `HETZNER_ARM64_HEALTHY=false` via ps57 Script Console for ~5 min, trigger again. Confirm an EC2 `m8g.2xlarge` provisions from the Graviton Fleet; `currentBuild.description` reads `[FALLBACK AWS:docker-32gb-aarch64]`. ccache writes to AWS S3, not Hetzner FSN1.
3. **Explicit Hetzner regression**: trigger with `CLOUD=Hetzner` while the flag is still forced unhealthy. Confirm the build stays on Hetzner regardless (no fallback by design).

#### Behaviour-change notes

- Unknown / empty `CLOUD` values: old code silently routed any unknown value to AWS. The new resolver `default:` case treats unknown as `auto` (Hetzner-when-healthy, AWS-on-fallback). Choice-param validation enforces the list, so this is only reachable via crafted API calls or the retry-trigger pass-through with empty `env.CLOUD`. Not a regression in practice; flagged for clarity.
- Direction-flip ccache warm-up: when `CLOUD=auto` flips between consecutive builds (Hetzner-healthy to AWS-fallback or back), the second build's ccache is empty in the target cloud and pays a fresh compile (bounded by 14d normal / 120d sanitizer TTL).
- Matrix-parent atomicity: `param-parallel-mtr.yml` passes `CLOUD` via `current-parameters: true`, so each matrix cell resolves independently against the controller flag. If the flag flips during fan-out, sibling leaves can land on different clouds. Desirable for fan-out resilience (one DC outage doesn't tank all siblings), not one-parent-decision atomicity.
- Retry-trigger at line 601 stays on `env.CLOUD` (not `CLOUD_CHOSEN`) so retries re-evaluate Hetzner health afresh instead of pinning to the fallback cloud picked on the first attempt.

#### Follow-ups not in this PR

- Queue timeout around the `node(LABEL)` block: today's pre-existing 6-day pipeline timeout means a healthy-at-resolver but unhealthy-during-provisioning race could queue on Hetzner for up to 6 days. Out of scope here (pre-existing risk; explicit-Hetzner callers face it today).
- `param-parallel-mtr.yml` adding `auto` to choices: hygiene only, since the parent's filter excludes all aarch64 cells today.

inikep (approved):
LGTM

PR #522, opened 2026-06-08
PS-10414: Set `CLOUD=auto` as default

inikep:


nogueiraanderson (approved):
LGTM. Correct and low-risk: the default matrix run is `x86_64` and stays on Hetzner, so `auto` only changes behavior for explicitly selected `aarch64`, where it adds the AWS Graviton fallback. Thanks!

nogueiraanderson (approved):

PR #524, opened 2026-06-16
PS-11214: Skip valgrind-only test list when present

inikep:
Add an optional --skip-test-list to MTR valgrind runs so tests known to time out or run too long under valgrind can be excluded without affecting normal/sanitizer runs. Guarded on the file's existence because mysql-test-run.pl errors out if --skip-test-list points to a missing file, so valgrind runs keep working until the list is shipped in mysql-test/collections/valgrind-skip.list.


percona-ysorokin (approved):
LGTM

PR #527, opened 2026-06-22
PS-11293 Update the pattern to match the 8.0 eol version format

VarunNagaraju:
https://perconadev.atlassian.net/browse/PS-11293

Also adds a case to match upstream 5.7.* versions as well.

### DIFF
--- a/jenkins/suites-groups.sh
+++ b/jenkins/suites-groups.sh
@@ -287,6 +287,12 @@ function set_suites() {
     5.7.*-*)
       set_suites_ps57 "$build_type"
       ;;
+    5.7.*)
+      set_suites_mysql57 "$build_type"
+      ;;
+    8.0.46-eol-*)
+      set_suites_mysql80 "$build_type"
+      ;;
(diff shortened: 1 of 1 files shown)


jenkins/suites-groups.sh
```diff
@@ -287,12 +287,12 @@ function set_suites() {
     5.7.*-*)
       set_suites_ps57 "$build_type"
       ;;
+    8.0.*|8.0.46-eol-*)
+      set_suites_mysql80 "$build_type"
+      ;;
```
inikep, line 295: This is wrong as `8.0.*` matches also `8.0.46-eol-*` and `8.0.*-*` in the next case.

inikep (changes requested):

jenkins/suites-groups.sh
VarunNagaraju: Fixed.

inikep (approved):
LGTM

